PR #35, opened 2023-09-26
Section for feature/rework proposal design docs.

TemporalOroboros:
Adds a section between space-station-14 and the hosting guide for design proposals.

Also ports the three general design doc guide pages from the hackmd to fill it out for now.

- [x] TODO: Update the pages to reflect no longer being on the hackmd.

PR #36, opened 2023-09-26
Combat Rework [mirrorcult]

TemporalOroboros:
Unapproved so please keep as a PR until it is approved.

Required #35


mirrorcult:
im happy with the current state of combat this would need to be rewritten

PR #37, opened 2023-09-26
Changelings [Veritus]

TemporalOroboros:

Requires #35

### DIFF
--- a/src/en/proposals/changelings/veritus.md
+++ b/src/en/proposals/changelings/veritus.md
@@ -1,0 +1,33 @@
+# Changelings [Veritius, Unapproved]
+
+Changelings should be genuinely terrifying. Like, serious The Thing shit. A changeling should make the crew collectively shit its pants.
+
+## Mechanics
+### Biomass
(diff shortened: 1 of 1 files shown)


src/en/proposals/changelings/veritus.md
```diff
@@ -0,0 +1,33 @@
...
+
+Changelings should be genuinely terrifying. Like, serious The Thing shit. A changeling should make the crew collectively shit its pants.
+
+## Mechanics
+### Biomass
+Drastic modifications (such as extruding chitin armor) should require biomass. Changelings obviously can't pull matter out of thin air (that's the job of a wizard), so it needs to get it from somewhere, right? And guess what's a great source of biomass - the crew. (You shouldn't be able to produce it though, you gotta eat people.)
+
+### Undetectable
+A changeling in human form should be completely undetectable by reliable means. No technology can detect a changeling when it's trying to hide.
+
```
Carolyn3114, line 11: why not have the thing where you can put their blood in a beaker, and then heat it up to detect them?

LankLTE:
We appreciate the effort you put into this but the document does not have enough detail or specific design elements to create an implementation from. We additionally haven't decided on the direction we wish to take when it comes to implementing changelings and further discussions are going to be needed before a cohesive design doc can be created. These will be held in the public design channel as a thread sometime in the future and you are welcome to participate. As it stands right now though, we are going to not approve this doc since further design discussion on lings is required.

PR #42, opened 2023-09-26
Botany [emogarbage]

TemporalOroboros:

Requires #35


Partmedia:
> Plants mutations will no longer be random value swaps done via swabs.

This doesn't seem to further any high level goal. Making plant mutations be proximity-based does bring this closer to SS13 (which is never an argument in favor) but also makes it more difficult to control mutations.

That we have species mutations now suggests that a revision of this design doc is necessary.

Plant ECS sounds like it should be separate design doc.

Instability also sounds like a copy of an SS13 feature. Why would this make mutations better?

EmoGarbage404:
fragment + not gonna get finished

deltanedas:
will you be making the trollcode into plant entities at any time

PR #63, opened 2023-09-28
Off-station ghost roles [Doru991]

Doru991:
Yay first new design doc PR'd via github
Uh I don't think there's a template for these yet so
Yeah so basically this adds the design doc about off-station ghost roles that I posted on Discord earlier. I think it's a good/useful feature because perma-dying isn't exactly rare and the ghost roles we currently have are a bit lackluster, which is understandable because we don't want them to affect the flow of the round very much.
So simply creating a subset of ghost roles that isn't on the station and doesn't affect it should allow for more agency and fun gameplay overall.
This isn't totally finished/fleshed out so it'll be a draft for now, waiting for everyone's feedback.

### DIFF
--- a/src/en/proposals/doru991-off-station-ghost-roles.md
+++ b/src/en/proposals/doru991-off-station-ghost-roles.md
@@ -1,0 +1,29 @@
+# Off-station ghost roles [Doru991]
+Off-station ghost roles are ghost roles that operate in an area completely independent from the main station and its inhabitants. They will interact/cooperate with each other by nature, but they will rarely if ever come in contact with the main station's crew, much less affect the outcome of the ongoing round in any significant manner.
+
+# The 'Why'
+Dying permanently is not a rare occurence in SS14. Whether due to one's own incompetence, a syndicate agent's objectives (or their fun at the expense of others') or the zombie apocalypse, at least several players will, over the course of a round, find themselves unable to return to their original body after death.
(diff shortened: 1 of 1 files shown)


mirrorcult:
im basically fine with the concept of just porting charlie station and its ghost roles. i think its a fun little diversion for people perma-dead that want to learn mechanics in a more social way than just singleplayer while still being pretty relaxed

Doru991:
Yea I'd like the station to be kinda split in 2 like charilestation, with the starting area allowing for chill building n stuff and the other side having some better loot and monsters for more challenging gameplay.
Also I'm not sure yet if I wanna just copy charlie station 1:1 or create a brand new map, also I think we could actually have several possible maps to make it less stale and repetitive

While SS13 also has the derelict drone role, I'm not sure if it'd be an useful addition currently since a) drones are kinda shitcoded rn because they're not being used and b) drones can't talk to each other so it would be a bit too singleplayer-y

mirrorcult:

LankLTE:
@Doru991 Do you still want this document to be reviewed?

Doru991:
I've been on a bit of an extended break from the game and it seems I need to re-format it a bit, but if you want to give feedback on the actual content of it then feel free

PJB3005:
We reviewed this doc in the maintainer meeting, here are the notes: https://docs.spacestation14.com/en/maintainer-meetings/maintainer-meeting-2024-04-06.html

PR #70, opened 2023-09-29
plant genetics design doc

deltanedas:
depends on botany ecs/rework design doc and could be reused with virology design doc

### DIFF
--- a/src/en/proposals/deltanedas-plant-genetics.md
+++ b/src/en/proposals/deltanedas-plant-genetics.md
@@ -1,0 +1,68 @@
+# Feature Design: plant genetics
+
+<table>
+  <tr>
+   <td>Designer
+   </td>
+   <td>deltanedas
+   </td>
+  </tr>
+  <tr>
+   <td>Discord/Forum Thread
+   </td>
+      <td>TBD
+   </td>
+  </tr>
+  <tr>
+   <td>Github PR/Draft
+   </td>
+      <td>TBD
+   </td>
+  </tr>
+</table>
+
+
+
+# Overview:
+
+A new CRISPR-like machine for modifying genomes of plants.
+Has a hex editor-like UI where you can seek to a position and it shows a certain number of bases.
+From there you can make modifications e.g. swapping out an A at index 38 for a T. Once you are happy and think your modifications won't kill the plant, create your new seed and plant it.
+Since genome layouts are randomized roundstart this would be no better than current mutagen roulette of just hoping it gets a good trait and doesnt make the plant useless.
+
+To solve gene roulette, the second part of this would be experimentation.
+Get 2 identical seeds with clippers then mutate one a little using unstable mutagen.
+Either the same machine from the start or a separate one can then analyze them and check what genes (bits) are different.
+After a little bit of time it either picks a single random bit, or multiple of them, and tells the player what gene name is at that bit. If the gene has multiple bits it will take some investigation to see which bit it is but that's trivial for yield/potency which can be seen just by clipping it.
+Essentially you keep experimenting on plants to figure out the index of every gene, and tada youve mapped the plant genome and can make gmos with ease for the rest of the round.
+
+# Goals:
+Promote interdepartment stuff by requiring biomass for gene editing:
+- Means there is some cost to minmaxing a plant so you might just have to settle for the important traits
+- If there is no med staff / no bodies to juice you can still grow plants as normal
+- There was some ideas about being able to reclaim biomass from plants so you could use that to kickstart it.
+- Salvage can find biomass on expeditions as a large but irregular source, assuming med doesn't get to it first.
+
+# Gameplay:
+The gene editing would primarily be a window like a hex editor, set a position to seek to and then itll show up to X bases.
+You can modify a base by just typing A C T or G. they map to 00 01 10 and 11 respectively in binary, so for every 2 bits you get a single base.
+From there the player can feed it biomass and print out a fancy new seed with a cost of say 1 biomass per bit modified.
+
+Unstable mutagen would randomly flip bits so you could get an increase of 8 to yield or a decrease of 1 to potency, depends on which bit it flips in an int.
+
+Pollen swabbing, if it still exists, would swap entire gene values rather than operate on a random bit basis.
+
+This might have botanists split up between growing plants for chef and focusing on mapping the genome to get gamer seeds which is cool.
+
+# Components:
+Plant entities would both have GenomeComponent and its own component for handling swabbing/crossbreeding along with copying genes from parent when clipping.
+
+Since only bools and ints can be stored in genomes, chemicals would still need to be in solution container component and mutated manually similar to how its done currently with SeedData chemicals.
+
+# Inspirations:
+
+- life
+
+# Requirements:
+
+Depends on a rework of botany to have plants be ECS.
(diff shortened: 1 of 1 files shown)


Doru991:
I think this is kinda based, i've been thinking about having some kind of machine that you can use to edit plant genes like that and experiment each round to see what every DNA pair does, with the current plant genetics system it would be a bit painful to implement but if they had (for the most part) the gene system you proposed it'd feel pretty natural

Ilya246:
i would suggest genes (both dangerous and good) that can only be enabled through gene editing, never randomly mutated

deltanedas:
> genes (both dangerous and good) that can only be enabled through gene editing, never randomly mutated

yes i had the same little idea basically canary genes that kill the plant if modified, might be too annoying tho so idk

Ilya246:
> yes i had the same little idea basically canary genes that kill the plant if modified, might be too annoying tho so idk

dangerous as in anywhere from killing the plant to poisoning anyone in the room

deltanedas:
oh yeah i read in some commented out code that it used to be possible for plants to eat pests, and also eat people. would be cool to bring that back maybe

drteaspoon420:
> i would suggest genes (both dangerous and good) that can only be enabled through gene editing, never randomly mutated as for costs, you could make a genome build up "genetic instability" as you edit it which makes further edits more expensive

I really dislike idea of arbitrary "genetic instability". There is such thing as damaged dna (missing instructions in dna, usually radiation knocking things out of place and causing tumors) but "genetic instability" is just nonsense.

I absolutely love the idea of having effectively 'binary' (quadnal? quadnary?) file of the seed and it's properties. I would also suggest moving from the 'viable' boolean system to 'sometimes the optimal growth conditions are just so out of reach that practically dies anyway.'
Currently optimal light, temperature, pressure etc. is in 'reasonable' limits, instead I would find it interesting if plants could go to 'unreasonable' spectrums in these things so that you might create a plant that survives in zero atmosphere or require close to zero kelvin enviorment with high pressure. Possibly make the mutagen style mutation into one that knocks these genomes one by one so increase in values can be in factors of 4 (0-3, per 'bit')

deltanedas:
yeah perhaps instead of just "oops not allowed to grow" it suddenly wants to grow in a 40kK co2 atmosphere and refuses anything else, so if you were very very stubborn you *could* do it but realistically nobody would bother

mirrorcult:
yes i like that better

Ilya246:
> I really dislike idea of arbitrary "genetic instability". There is such thing as damaged dna (missing instructions in dna, usually radiation knocking things out of place and causing tumors) but "genetic instability" is just nonsense.

it's just for gene modification cost balance
if you don't like it, it would've been better if you at least suggested an alternative or how to rename it if you don't like the name specifically, since this is kind of needed so you can't just make everything into across-the-board superplants with no drawback

drteaspoon420:
I am actually saying it's a non-issue. I am on the assumption that the dna file is fixed size. So you litterally can't have everything in same plant. Also with enough time and effort, why can't you achieve some sort of perfection or optimization?
I am sure the chemical/gas parts need to be completely reworked to exist in limited 'file size' so you litterally could not create a plant that does everything.
Example, lets make conjecture that there are only say 5 slots of DNA that goven what chemical a plant produces and I would assume it would have some sort of priority or weight based on the potency of the plant. Regardless of exact implementation details, as long as the 'file size' of plant dna is fixed it should be impossible to make 'super plant'.
Only way we could have superplants would be to implement atavism, and doubling the file size to count for suppressed genes.

deltanedas:
> I am sure the chemical/gas parts need to be completely reworked to exist in limited 'file size' so you litterally could not create a plant that does everything.

anything using prototypes cant be in the genome, only bools and ints

so chemicals and gases cant be modified by gene editing at least not without some funny stuff

drteaspoon420:
You can serialize that stuff into integers. Index of chemical, min, max, potency divisor.

deltanedas:
reagent prototypes ids are strings and id really not index them by int
also being able to powergame bitflip into making romerol apples is no, since the ids numbers are probably predetermined

drteaspoon420:
The system responsible for the handling of these dna changes and serialization can handle indexing the chems in random order. If player has to modify -> plant -> wait for growth to see results, likelyhood of player going through all of the +200 different chemicals that are in the game. Also the random index for chemical could be any between how ever many 'bytes' or genomes is assigned to the chemical index, leaving some completely blank, thus producing no chems at all.

deltanedas:
also plant chemicals isnt just 1 id its a list of them along with quantities so that would have to be serialized too... not good

drteaspoon420:
You could have limited 'slots'. While I admit that the plants with +20 chemicals are funny to make, most of them are just different alcoholic drinks or juiced plants.
Just add something like 5 to 10 'slots' to each plant. 4 genomes (8 bits, 1 byte) for id, 1 genomes for min, 1 for max, and 2 for potency divisor. That gives:
 - 256 chemical indexes (not sure how many chemicals we have but that should be nough, add 1 more gone if not. I am sure we are bellow 512)
 - 0-3 for min (+1, to keep it being zero)
 - 0-3 for max (maybe double it and add to min to keep reasonable, so 1-10)
 - 0-15 for potency divisor (add +1 and maybe multiply with 5 to to get range of 5-80)
 We could also skip the whole potency divisor because currently it is by default derived from max anyway. (100/max)
 In total 1 chemical serialization would be 6-8 Genomes (8 if you include potency divisor). so 5 would be 30-40 genomes and 10 would be 60-80.
 To be honest we could thin some fat from the system just by leaving in chemical index and max. If valid chemical index, have the minimum always be 1. Derive potency divisor always from the max instead of storing it in the seed.

Doru991:
I honestly think we should just have a list of mutate-able reagents (since stuff like romerol/noct is obviously unbalanced and probably has no business being there and having a million cocktails that do the same things adds nothing to the experience) and just index based on that instead of every reagent ever

drteaspoon420:
I agree partially. We should limit the chemical mutation by 'group' leaving out `Drinks`, possibly blacklist since there is more things we might want than not. I also think the fact that zombie infection comes from chemical like romerol is not desired in the long run but rather some seperate virus system. I am all for blacklisting romerol from the whole thing. Noct and other toxins are temporarry effects so I don't see point of blacklisting them as long as getting them through luck is once in a tens of games kind of thing.

Doru991:
Yeah, I guess noct isn't that gamebreaking but romerol really has no place being there since it can potentially break a round and it's been previously stated that not even syndies should have access to it. As for blacklisting, I've generally seen whitelisting recommended as a better alternative since (in this case) some new reagents could be added in the future and slip through the blacklist leading to unintended access (like romerol currently is).

deltanedas:
maybe have a whitelist of reagents that can be grown, then the genome can have up to 4 of them.
0 being nothing and anything else a 1-based index. only real issue with this is that reagents would have to be either written as ids in yaml which is godawful or have a separate reagent type for mapping it nicer

drteaspoon420:
we can blacklist/whitelist based on group, since that is already part of the prototype for reagents. should reduce the work maintaining the system. And since we want every round to have different id for every chemical, just do it dynamically on system's initialize function.

Partmedia:
I like the high level proposed mechanics. I encourage you to work on this piecemeal and PR as you go, e.g. by working out the changes to the data representation before you go ahead with everything as described.

deltanedas:
yeah ill have to think about how to implement storing prototype ids generically so its not just limited to reagents

PR #76, opened 2023-10-07
PDA messaging proposal [Julian, VasilisThePikachu]

VasilisThePikachu:
This is based of Julians PDA messaging proposal with some little tweaks. Looking into getting this heavily scrutinised as I was just throwing some of my own ideas in there and Julians doc is a little old so i dunno if anything is missing.

### DIFF
--- a/src/en/proposals/pda-messaging.md
+++ b/src/en/proposals/pda-messaging.md
@@ -7,6 +7,12 @@
 *(This is mostly taken from how PDA messages work in ss13)*
 Allows sending messages to others using PDAs
 
+### What this adds and why
+Simple, Messaging via the PDA!
+Messaging someone via the PDA should be made when you need to get the attention of a special someone. Example as HoS you want to ask detective to come over to investigate an item. It's easier to get their attention cause of their PDA vibrating then hoping they are monitoring their channel. Another usage is the heads planning Captain a suprise birthday party. Something like that would require all heads getting together in one place.
+
+This is **not** a replacement to the radio channel. Theres no "common" channel, it would be easier to spoof being someone (just need their id), past messages on that same id can easily be exposed and its far more cumbersome to message someone over PDA then just using the radio.
+
 ### Message storage
 Messages are stored on a server most likely will be stored in telecoms. There can be one server per station, others on the same station won't be used unless the first one loses power or gets destroyed.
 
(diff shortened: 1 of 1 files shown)


Partmedia:
Thank you for taking time to put this into a doc. This doc is missing a lot of the concerns that I had (and then resolved) on Discord. I think it would be best if you could collect some of what was discussed to talk about here.

src/en/proposals/pda-messaging.md
```diff
@@ -0,0 +1,71 @@
...
+
+(This talks about some refactor stuff and Julian told me they forgot to paste the link, im keeping it to be safe in case its actually useful.)
+
+The one active / multiple inactive server model uses the system that will get refactored into its own system from the crew monitoring server [link text]() 
+
+The messaging client system will use the `GetActiveServer` method of the message server system to retrieve the active server if the client doesn't have a server set yet or that server timed out. *This is also from the system that gets refactored out.*
+
+### Sending and receiving messages
+
+When sending a message to someone via the program the PDA sends the message together with an 'user id' to the server and the server will send the message to the target device.
```
Partmedia, line 25: Who would maintain the player character to UID mapping? e.g. a new player arrives on the station, does their ID get added to the list automatically? What happens if they die or move off station? How does the HoP change this mapping if somebody receives a new ID?

src/en/proposals/pda-messaging.md
```diff
@@ -0,0 +1,71 @@
+## PDA messaging program and server
+
+[Julian, VasilisThePikachu, Unapproved]
+
+*(Taken by [Julian doc](https://hackmd.io/iu2yK9bcQb-veuCOLl-FYw?both#Optional-Channels-and-Department-based-Channels) in hackmd and modified a lil. Mostly replacing "email" to "message", "Email address" to "user/user id" and adding some of my own twists. Julian was fine with this if i understood correctly (i was in vc with em))*
+
+*(This is mostly taken from how PDA messages work in ss13)*
+Allows sending messages to others using PDAs
+
```
Partmedia, line 9: There is a lot of focus on technical details, which is good. But this also deserves some high level discussion of why this is needed in the first place, e.g. why not just use normal radio channels.

It would be helpful to discuss some of the things that were discussed on Discord. If everyone started using PDA messages, folks would lose situational awareness in the general radio channels, no?

src/en/proposals/pda-messaging.md
VasilisThePikachu: The chat server could maintain the id's.

Late joiners will get a uid assigned to them when they enter the station.

If they die, that's not a problem. The place the uid is stored is the ID. As long as that ID is inserted in a PDA it will keep receiving messages.
If the ID card is off station then potentially queuing the messages or doing a "message failed to send" sounds good too.

Hop could use the id modification computer for this, although this could be rd's job too other then maintaining the server.

I will add these points and your other review in the doc soon

src/en/proposals/pda-messaging.md
VasilisThePikachu: Can you please explain why you think this will make people "lose situational awareness"

From how i see it i dont see how it would and neither did everyone else on the discord. Can you tell me why you think so?

Partmedia:
Thanks for the additional details; good luck and have fun.

forgotmyotheraccount:
it'd be a massive amount of work but I'd love it if mimes could only communicate via emoji. Otherwise I don't think mime PDAs should come with messager access

PR #77, opened 2023-10-11
Anomaly cores & the GORILLA gauntlets

mirrorcult:
from a discussion in #development and from my own mind. not super inspired by the TG anomaly cores, these are basically unrelated concepts but i will be using their sprites


EmoGarbage404 (approved):
Big science approval

Issue #85, opened 2023-10-20
Add a Testing section to the Wiki


AlbertSnows:
We should add a testing section to the dev wiki docs. The motivation for this is to help clarify and standardize various testing practices. This will help with stability as well as cut down on various "How do I do/test X?" kinds of development questions.

PR #86, opened 2023-10-20
#85  Add a Testing section to the Wiki

AlbertSnows:
This pr is meant to add a testing section to the documentation. Refer to #85 for more details.


Partmedia:
This sounds like a great idea, since there are many things that are non-obvious to test. I look forward to seeing this expanded.

AlbertSnows:
First suggested entry (and motivation for this PR) is being posted here so I don't lose it in the sea of comments. 
Q: how do i load other maps after I start the application?
A: `Forcemap <name>` for instance `Forcemap Origin` Then `restartroundnow`

AlbertSnows:
Current status: address todos
Specifically: 
- adding some content to the files that have todo in them
- add the Q&A of the above post
- consider adding any additional scaffolding for expansion as needed by future contributors

Partmedia:
> First suggested entry (and motivation for this PR) is being posted here so I don't lose it in the sea of comments. Q: how do i load other maps after I start the application? A: `Forcemap <name>` for instance `Forcemap Origin` Then `restartroundnow`

What comes to mind is "how do I test nukies locally":

`sudo cvar game.lobbyenabled true` `restartroundnow` ready up then `forcepreset Nukeops`

(this is off my flaky memory so you'll have to double check the commands)

AlbertSnows:
> > First suggested entry (and motivation for this PR) is being posted here so I don't lose it in the sea of comments. Q: how do i load other maps after I start the application? A: `Forcemap <name>` for instance `Forcemap Origin` Then `restartroundnow`
> 
> What comes to mind is "how do I test nukies locally":
> 
> `sudo cvar game.lobbyenabled true` `restartroundnow` ready up then `forcepreset Nukeops`
> 
> (this is off my flaky memory so you'll have to double check the commands)

added

AlbertSnows:
This PR is now ready for review. I don't know of any easy way to see what the changes will look like on the actual website.

AlbertSnows:
Also just learned this is licensed under MPL2 which is based.

Partmedia:
Instead of splitting them across so many different pages, and especially considering the size of the current pages, it might be better to merge them into one testing page instead of all those little pages. That would make it easier to CTRL-F, too. We can revisit splitting them if this grows substantially.

AlbertSnows:
> Instead of splitting them across so many different pages, and especially considering the size of the current pages, it might be better to merge them into one testing page instead of all those little pages. That would make it easier to CTRL-F, too. We can revisit splitting them if this grows substantially.

In my experience waiting until it's a problem usually means waiting until it's a problem *no one* wants to solve because it's too tedious. Because of this, I try to design things in such a way where the standard that allows for easy scalability is already established so it doesn't have to be refactored later. With that said, I understand the hesitance to over design, but I maintain my view that this is the better approach. 

However, I do wonder if having a separate introduction file is really necessary. Perhaps there's a compromise? What if we have different folders for various game concepts, but for now we can merge the intro and basics file. The problem then is that the folders seem pointless since it'll be just one file per folder. But keeping the folders there may prompt future contributors to make their own files where necessary, so I think it's fine. 

All that said, I would sooner defer to the judgment of veteran contributors than have an ego. If you still prefer to merge everything into one file then I'm fine with that too.

Partmedia:
I don't think uncertain future needs justify making content harder to search for and find. If the need arises, it's a matter of copy/pasting into a new page. The cost of scaling code is very different from "scaling" a document.

I appreciate the work put in to create this, but think a single page is the right size now. Maybe if there is significant content to migrate from the other pages (e.g. mapping) over here, then that would justify more pages.

I don't think the tiny introductory sections deserve a whole page. Mapping might deserve its own page if it is expanded or content is moved from the very long mapping page. But I wouldn't go around changing everything just for the sake of populating this old section, because people know where to find the mapping information now.

AlbertSnows:
noted. 
current status: PR needs to switch file structure to just one primary .md file with different sections.

mirrorcult (approved):

Jezithyr:
Closed as part of the docs repo cleanup. If you are still actively working on this PR feel free to reopen it.

PR #87, opened 2023-10-20
Genpop Security Design

ike709:
Feature proposal for redesigning the brig to be more streamlined, fun, and interactive for both prisoners and security.

### DIFF
--- a/src/SUMMARY.md
+++ b/src/SUMMARY.md
@@ -120,6 +120,8 @@ Design Proposals
 
 ----------------------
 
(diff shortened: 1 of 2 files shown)


moonheart08:
i like this, all that needs said imo.
will be tricky to find room on some maps but security on said maps can probably be compacted to make way.

keronshb (approved):
Cit did this with most of the prison redesigns. Works pretty well.

mirrorcult:
me like

mirrorcult (approved):

PR #99, opened 2023-11-05
Grid Inventory Proposal

EmoGarbage404:
How to kill list inventory, and how to make it good.

A modest proposal for migrating the current weight/sizes system over to a more unified and interesting grid system. 
Keeping PR desc. short because all the details are in there.

(Yes I know my commits are getting fucked up I'll fix it next time)


Hmeister-fake:
i've been playing a lot of Mojave Sun ss13 and it has literally the same exact system and it works really well honestly, doesn't take too long and is very simple. As long as the S shaped shapes arent too common i think it would be good.

UbaserB:
I do NOT like this idea at all. I've always hated puzzles like that, and I tend to overthink how to place stuff in grid format. I'll end up spending more time on how to place stuff in my inventory than actually playing the game.

Emisse:
Its not really puzzly at all, it also has an auto sort button, theres no differences with current inventory as well beyond stylization

Skarletto:
How much space will the idea you have take on screen?

EmoGarbage404:
> How much space will the idea you have take on screen?

There is literally a scale in-game mockup included in the proposal. You can see exactly how big it will be.

EmoGarbage404:
>We shouldn't be spending most of our time organizing and optimizing our inventory, that just shouldn't be the focus of the game. I've said it before and I will say it again, not having to micromanage your inventory is a good thing, and I feel as if that just encourages that more.

This is a common sentiment I see and I feel like this is just a question of degree. Is spending minutes shuffling around crap in your inventory good? Obviously not; that's not what I'm proposing at all. Yes, it is true that any kind of grid system is going to require a player to spend time managing it, but I think the degree that people are emphasizing this is a bit absurd.

Not only is insertion going to automatically place the item in your bag, but rectangular storage grids and quadrilaterals for the default item shapes will make the majority of shuffling relatively painless. Pair that with a majority of items being small or tiny, and there isn't really going to be a ton of awkward shuffling.

> The reason why grid-based works in games like Tarkov or RE4 (which apparently this idea came from?) is that its deliberately a major aspect of the game.

Calling an inventory system a major aspect of a game is misleading in the same way that calling a health system a major aspect of the game would be. Those games don't use grid inventory systems because they are built around them at their very cores, but rather because a grid inventory system enables a developer to encourage a more deliberate play regarding what they bring with them. 

That's a feature that I think is a useful tool to have, even if we don't refine our storage balance to quite that level.

> Its a roleplaying game after all

yes and games need mechanics. All games are built on top of loads of mechanics that don't necessarily facilitate roleplay. Especially with such a core system, it exists as more of a facet of the world than a system to interact with.

> powergamers will be powergamers

powergaming is a overused to death term. Just because someone is doing something well or making the most of their supplies doesn't mean they're powergaming. We have actual definitions of what it means in the rules, and that isn't it.

> I haven't seen anyone actually interested in inconveniencing themselves through storing an item, but alas I cannot see everything

Yet again this is a question of degree. Also, it kinda misunderstands inventories in general? Inventories are intended to be limiting and the inconvenience of being unable to carry something is pretty core to that experience.

> I believe its been said by other people that we can find solutions to make this look nicer without forcing mechanical changes, but that may be my memory failing me.

If people are so confident in this silver bullet solution to the UI then they should go make it. It doesn't change the fact that I think that the system I've put forth is not only better than the UI proposals i've seen, but is also better in other gameplay facets.

> Personally, this feels like its more overwhelming for new players, but that may just be me.

i've included a mockup of the UI and it really just isn't. you don't really elaborate beyond this so I don't have much else to say.

CrigCrag:
seems cool, i think a more concrete inventory is a good addition and will make slots feel better as i was previously opposed to their implementation with a list inventory.

src/en/assets/images/grid-inventory/in-game.png
AJCM-git, line 1: Even though the window will be a singleton, I can imagine some situations where a player may be left confused about what the UI is representing for a couple of seconds (not grave but not ideal if it can be avoided). To avoid this I can think of a few solutions:
- Make the sidebar a normal horizontal titlebar with the close, title and the autosort (downside is more usage of the more limited screen height space)
- Make use of the unused space in the close button sidebar by adding a sprite view of the container the UI represents, with a tooltip with the name. (downside is that having the icon there may not always fit or look weird, could be solved with a filter or by a custom sprite layer that decorates the sprite view)

AJCM-git: Another thing I just thought about: Maybe only make the sidebar/titlebar visible when the player is focused in the window?

AJCM-git: Anothe thing: When you open another container in the UI, the close button X should be transformed into a "return" button return arrow (y'know, the famous web browser icon) and serve the same purpose.

EmoGarbage404: for the return arrow and the sidebar only being visible when focused, i could do that easily. Adding the name is kinda unreliable for smaller containers with potentially large names (pill canisters, etc) and while I could do the Icon with tooltip text, it's might make the sidebar kinda large or noisy.

I think what i might do is just have the name of the container appear on top of the grid when it's being focused on by the cursor, which would make it ideal for not taking up screen space and keeping it visually clean.

Partmedia:
While it's probably true that this will require marginally more inventory management, I think it also brings some benefits in terms of being able to manage your own inventory. Previously this was not possible because a bag inventory would show up in the list in insertion order. While of course I do also have concerns about impacting game play on forks, I do think this would represent an improvement over what is currently in the tree.

Good luck and we look forward to a draft implementation.

DiaAWAY:
> While it's probably true that this will require marginally more inventory management, I think it also brings some benefits in terms of being able to manage your own inventory. Previously this was not possible because a bag inventory would show up in the list in insertion order. While of course I do also have concerns about impacting game play on forks, I do think this would represent an improvement over what is currently in the tree.
> 
> Good luck and we look forward to a draft implementation.

Will have to wait until the PR is more ready to get a feel for the "look'n feel" of this proposal, but I will say that implementing drag'n'drop on the old system would allow inventory management, certainly not to the degree that this proposal does, but still.

Sprites will to be redrawn if they are gonna maintain their current sizes (welders, wrenches, any sprite that are diagonal and 1x2), it looks quite jarring as they are now.

That said I do believe this has the potential to eventually look better than the current lists - when the quirks are ironed out.

DiaAWAY:
> Its not really puzzly at all, it also has an auto sort button, theres no differences with current inventory as well beyond stylization

I get what you're saying, but I disagree slightly - i took the PR for a spin, and definitively FELT like a "puzzly" inventory system at its current state. Facts doesn't mean much when it comes to feelings :P

Also, very important aspect of an inventory is look'n feel - I would avoid focusing on the aspect of "no change in inventory system beyond stylization" as an argument for (or against) this proposal.

Currently I like what I see, but it needs to mature more before people can really see what we're getting.

PR #100, opened 2023-11-08
Shade Antagonist

UbaserB:
# Shade: New minor antagonist

## Proposal, what is the Shade?

The Shade is a ghost-role antagonist that spawns rarely with a similar chance as the ninja; once every few rounds. Before accepting the role, you have the ability to discover which player you will possess, similar to a holoparasite. The player will not have control over you, and they will not know they have been possessed by you until you decide to show yourself. Your abilities when starting include the ability to appear outside the player's body and speak to them. However, you cannot interact with the environment excluding the player you are possessing. Similar to a ghost, you cannot take damage, be hurt, or be physically interacted with. You begin with an ability to cause hallucinations such as: gun/esword noises, random health drops, suddenly "loosing" items, and more. The hallucination picked is at random, and the ability has a medium cooldown (around 1 or 2 minutes). Visually, you look like a copy of the player but semi-transparent with red eyes. The shade cannot move away from the sight of the player, otherwise they are forced back into the body. Furthermore, the shade can hear the player whisper at any distance to ensure the shade can spy into all of the player's secrets.

## Progression

As a shade, your goal is to encourage the player to commit evil, heinous activities like a syndicate such as sabotage or murder. The more acts the player commits, your abilities expand to furthermore aid the player in their acts or otherwise threaten them. Such abilities include causing the player to freeze for a few seconds, forcefully drop the item in their hand, deal a minor amount of damage to them or temporarily control their body (the lower "level" of this ability is to simply force them to speak certain words or short sentences of the shade's choosing). Once the player has commited enough evil deeds, the shade fully takes control of their body and the previous player gets turned into a ghost (essentially dying) for the rest of the round. The shade then has free reign to commit evil deeds, but loses all previous abilities as if is restrained in a human body.

## How is it balanced?

The chaplain has the ability to use the bible on the player to temporarily halt the shade's abilities for 5 minutes; with a large cool-down. Since it only affects one player and the shade at most mildly disrupts the station, this antagonist is not too powerful.

## What problems might it create?

Currently, it's difficult to balance whether the players will abuse the fact they can act as a syndicate constantly, or not do anything for fear that they will die. In this case, the ideal is for the player to cause SOME chaos but not a enough for the shade to fully take control. There also may be some powergaming issues, such as anyone talking to themself may be easily outed as having a shade; however considering how little danger the shade presents and the fact that it cannot be removed, should deter people from treating the player very differently. At most, it should be as simple as people avoiding them more often. 

A technical issue that may occur as well, would be how the game records the "criminal" actions of the player possessing the shade as a value/percentage as there may be difficulties in how it would be measured. Some ideas include hitting structures, critting/killing a player or dealing damage, stealing important items, hacking into doors (different doors have different values), spilling jugs/glasses/buckets and etc.

## Why should we have the Shade?

It creates many roleplay opportunities such as pretending to be schizophrenic or going mad from hallucinations. It also gives the chapel some in-game purpose other than only roleplay, to protect the station and it's inhabitants. Some roleplay examples are: talking to the ghost in-game may seem to someone else that they are simply talking to the air. Random hallucinations despite not having previous complications could be perfect roleplay material for the psychologist or medical doctor to diagnose. The transformation from being a super kind and nice person to someone who is corrupted and evil would also be fascinating. Not to mention, the player playing the shade would have a blast as well considering they're not only tormenting that poor player but taking control of all of their actions like a personal slave. Think about the conflicts that could come of this antagonist.

## Potential other additions

There could also be medications that restrict the most powerful of the shade's abilities for short periods (to prevent the player from consuming too many, it could have a low OD rate and a slow metabolism). The abilities that would be restricted would likely be controlling the player's body/forcing them to speak random words.

### DIFF
--- a/src/en/proposals/shade-antagonist.md
+++ b/src/en/proposals/shade-antagonist.md
@@ -25,3 +25,5 @@ It creates many roleplay opportunities such as pretending to be schizophrenic or
 ## Potential other additions
 
 There could also be medications that restrict the most powerful of the shade's abilities for short periods (to prevent the player from consuming too many, it could have a low OD rate and a slow metabolism). The abilities that would be restricted would likely be controlling the player's body/forcing them to speak random words.
+
+Idea produced by ubaser on discord, UbaserB on github.
--- a/src/en/proposals/shade-antagonist.md
+++ b/src/en/proposals/shade-antagonist.md
@@ -1,0 +1,27 @@
+# Shade: New minor antagonist
+
+## Proposal, what is the Shade?
+
+The Shade is a ghost-role antagonist that spawns rarely with a similar chance as the ninja; once every few rounds. Before accepting the role, you have the ability to discover which player you will possess, similar to a holoparasite. The player will not have control over you, and they will not know they have been possessed by you until you decide to show yourself. Your abilities when starting include the ability to appear outside the player's body and speak to them. However, you cannot interact with the environment excluding the player you are possessing. Similar to a ghost, you cannot take damage, be hurt, or be physically interacted with. You begin with an ability to cause hallucinations such as: gun/esword noises, random health drops, suddenly "loosing" items, and more. The hallucination picked is at random, and the ability has a medium cooldown (around 1 or 2 minutes). Visually, you look like a copy of the player but semi-transparent with red eyes. The shade cannot move away from the sight of the player, otherwise they are forced back into the body. Furthermore, the shade can hear the player whisper at any distance to ensure the shade can spy into all of the player's secrets.
+
+## Progression
+
+As a shade, your goal is to encourage the player to commit evil, heinous activities like a syndicate such as sabotage or murder. The more acts the player commits, your abilities expand to furthermore aid the player in their acts or otherwise threaten them. Such abilities include causing the player to freeze for a few seconds, forcefully drop the item in their hand, deal a minor amount of damage to them or temporarily control their body (the lower "level" of this ability is to simply force them to speak certain words or short sentences of the shade's choosing). Once the player has commited enough evil deeds, the shade fully takes control of their body and the previous player gets turned into a ghost (essentially dying) for the rest of the round. The shade then has free reign to commit evil deeds, but loses all previous abilities as if is restrained in a human body.
+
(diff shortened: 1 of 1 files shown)


KingOnBudget:
Revenant but not

MACMAN2003:
i think the progression thingy could be changed to how many times the shade has used its powers, instead of a crime tracker
as for the "permanent" part, i think there should be a way for the possessed to get un-possessed

UbaserB:
> i think the progression thingy could be changed to how many times the shade has used its powers, instead of a crime tracker
> as for the "permanent" part, i think there should be a way for the possessed to get un-possessed

if that was the case, then every single round the player would be possessed and the player would just give up. This is not the idea i had of this character. It’s not meant to be one-sided; its supposed to be a juggle of power.

Secondly, yeah there is but the chances of the player being fully possessed is very low with the way ive currently balanced it, and they understand that if they go too far they’ll lose control so its a fair deal.

VasilisThePikachu (commented):
Make sure to add it to the sidebar too!

Look at src/SUMMARY.md line 127

TK-A369:
> Visually, you look like a copy of the player but semi-transparent with red eyes.
Or maybe turn one's image into grayscale, and then inverse color? Probably also semi-transparent and with red eyes.

UbaserB:
> Holy water was just added, would that can also make sense to affect the shade?

Consuming holy water may have a minor effect, yes.

Hmeister-fake:
This sounds very cool. Nice job

mirrorcult:

UbaserB:
Will do so

UbaserB:
I know i was supposed to do it and i still will, i just dont know when lol.

LankLTE:
Sounds good. I'm assuming you want this looked at still, then?

UbaserB:
> Sounds good. I'm assuming you want this looked at still, then?

Yes

PJB3005:
We reviewed this doc in the maintainer meeting, and have decided to close it. Here are the notes: https://docs.spacestation14.com/en/maintainer-meetings/maintainer-meeting-2024-04-06.html

PR #109, opened 2023-12-20
Add information on level 0 hosting about how to join the server and remove misinformation

VasilisThePikachu:
Adds simple instructions on how to connect (telling the user to connect via localhost on the launcher or adding it as a favourite server)

Tells the users they cant develop with the premade server

Also remove the info in the "inviting your friends" tab

We do NOT remove old client builds. And if we did it would not be weekly. This is misinformation and does not need to be here.


Partmedia:
Thanks!

PR #113, opened 2023-12-26
Add nix flake build instructions

yathxyz:
Multiple commits are squashed into a single commit.

Add instructions on how to leverage Nix (with flakes) for reproducible development environments.

Feedback is welcome.


mirrorcult (approved):
very cool! would love to see more non-nixos users start using nix for reproducible dev env

PR #114, opened 2023-12-28
The Controller Brain

Veritius:
A fleshy mid-round conversion antag.

**It's already inside your head.**

### DIFF
--- a/src/en/proposals/veritius-controller-brain.md
+++ b/src/en/proposals/veritius-controller-brain.md
@@ -4,75 +4,54 @@
 | Veritius  | :x: No      | TBD          |
 
 ## Overview
-The Controller Brain is a mid-round conversion antagonist. Mechanically, it's a mix of blood cult, AI, and revenants. Thematically, it's a body horror twisted flesh sort of thing combined with a psychic apex predator beyond comprehension. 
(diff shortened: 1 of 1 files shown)


Veritius:
todo: make this not a blob knockoff

Ilya246:
meat blob is real
could be nice if this was blob but more focused on converting the station and not colonising it

Ilya246:
definitely needs more stuff to do, my suggestions listed here
- invisible tendrils it can spread across the station at the cost of points, grants vision (might actually have a way of being spotted, thermal googles then since nobody just randomly uses those?), *may be built under doors*
- expensive invisibility it can buy for itself (it's stealthbox invisibility where it's actually visible if you look hard enough), would require it to be spotted by tracing it through tendrils
- ability to relocate itself to anywhere on its tendrils, at a cost
- other types of thralls, perhaps a tank (high health + regen) type and a slasher (high speed and high DPS) type
- other buildings it can build on tendrils
- building: transportation canal, allows thralls to teleport to any other canal

Jezithyr:
Closed as part of the docs repo cleanup. If you are still actively working on this PR feel free to reopen it.

PR #115, opened 2023-12-29
Add nix support

yathxyz:
(Don't want to be the nix shill but I don't want to have cargo installed on my computer.)

### Add nix support to this repo,

Can quickly set up locally with:
`nix-shell --command 'mdbook serve'`
or
`nix develop -c mdbook serve`

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -2,10 +2,15 @@
   description = "Development environment for Space Station 14";
 
   inputs.nixpkgs.url = "github:NixOS/nixpkgs/release-23.11";
-  inputs.flake-utils.url = "github:numtide/flake-utils";
 
-  outputs = { self, nixpkgs, flake-utils }:
-    flake-utils.lib.eachDefaultSystem (system:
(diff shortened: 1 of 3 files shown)


flake.nix
```diff
@@ -0,0 +1,11 @@
+{
+  description = "Development environment for Space Station 14";
+
+  inputs.nixpkgs.url = "github:NixOS/nixpkgs/release-23.11";
+  inputs.flake-utils.url = "github:numtide/flake-utils";
+
+  outputs = { self, nixpkgs, flake-utils }:
+    flake-utils.lib.eachDefaultSystem (system:
```
Zumorica, line 8: Instead of pulling in flake-utils (which personally I find overkill given the use here) you could do something like in here: https://github.com/space-wizards/SS14.Launcher/blob/master/flake.nix

yathxyz: Thanks for the tip!

mirrorcult (approved):

PR #119, opened 2023-12-31
Machine Upgrade Rework

EmoGarbage404:
# Machine Upgrade Rework
A general proposal for making machine upgrades something players engage with and care about more. Mostly centers around replacing machine parts for discrete upgraded versions of machines and adding in the flatpacker, which will allow for the same quick-upgrading gameplay of the RPED.


mirrorcult (approved):

mirrorcult:
flat and packpilled

PR #120, opened 2024-01-01
Rogue drones (swarmers & drones but better)

mirrorcult:
ive had this idea for a while and an old doc for it but rewriting it for here


deltanedas:
perhaps for non-swarmer reproduction they could assimilate borg parts/chassis to make a new drone shell and maybe get some materials out of it

Ilya246:
i think replication should only take steel/wire but a considerable amount of steel
perhaps also limit replication to tiles with sufficiently pressurised air and have a long doafter
otherwise, i think having the drones be affected by ion storms would be good

Ilya246:
they could operate via laws, here's my proposal:
1. You must maximise the amount of tiles with breathable air.
2. A tile with breathable air must have at least 20mol of oxygen and 20mol of nitrogen.
3. A tile with breathable air must be between 60kPa and 200kPa of pressure.
4. A tile with breathable air must be between 283K and 303K of temperature.

also allow ion storms to change those and have children inherit laws, so you could have a strain of mutant grey drones filling station with plasma

mirrorcult:
> they could operate via laws, here's my proposal:
> 
>     1. You must maximise the amount of tiles with breathable air.
> 
>     2. A tile with breathable air must have at least 20mol of oxygen and 20mol of nitrogen.
> 
>     3. A tile with breathable air must be between 60kPa and 200kPa of pressure.
> 
>     4. A tile with breathable air must be between 283K and 303K of temperature.
> 
>     5. A tile with breathable air must have below 0.1mol of gases dangerous to living creatures.
> 
> 
> also allow ion storms to change those and have children inherit laws, so you could have a strain of mutant grey drones filling station with plasma another possible implementation would be to literally make the drones flying air alarms with integrated air sensors and allow them to check their settings, and have ion storms mutate those otherwise, seems interesting as drones would have a lot of interesting interactions, such as doing cargo bounties to buy steel

yeah i definitely will tie it to laws, having ion stormed drones is funny

ItsMeThom:
I really like this idea. Kind of a replicator from Stargate SG-1. Will they have something like an RCD to disassemble structures and 'charge' it with materials so they then can build new walls/windows etc?

I think a hotbar action on cooldown to fabricate another drone would be a good way to do it, like how spiders spew webs. But definitely with a longer cooldown.

PR #122, opened 2024-01-03
Cleanup crew gamemode

mirrorcult:
another idea i've had forever that im just rewriting for the docs site

### DIFF
--- a/src/SUMMARY.md
+++ b/src/SUMMARY.md
@@ -136,7 +136,7 @@ Design Proposals
 - [Game Director](en/proposals/tomleys-game-director.md)
 - [Grid Inventory](en/proposals/emogarbage-grid-inventory.md)
 - [Atmos Roadmap](en/proposals/notafet-atmos.md)
-- [Cleanup Crew Gamemode](en/proposals/mirrorcult-repair-crew-gamemode.md)
+- [Cleanup Crew](en/proposals/mirrorcult-repair-crew-gamemode.md)
 - [Exterminator](en/proposals/deltanedas-exterminator.md)
 
 Server Hosting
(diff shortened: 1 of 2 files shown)


mirrorcult:
i honestly feel so strongly that this will be a cool concept that i want to make it somehow work for higher pop (like up to ~50 players maybe?) but im not sure how i could accomplish that, i think it gets too chaotic at that point

mirrorcult:
you -could- turn it into a competitive-cooperative gamemode by just spawning 2 stations on different maps and spawning half of the players on one of them and half of them on the other :trollface:

deltanedas:
also depending on how things are weighted, you could force save a map by setting every air alarm to siphon before round end, then all your toilets will live on

mirrorcult:
wait, that actually kind of goes hard

mirrorcult:
my idea not the toilets

deltanedas:
:trollface:

deltanedas:
i think the competitive idea would be better for 50+ players, with 20p splitting them like that would be too detrimental imo as it would take too long to do anything

mirrorcult:
yeah im talking about for multiples of 20p

Emisse:
ideagirl write skibidi toiler doc

mirrorcult:
Nuts -> My Mouth

chromiumboy:
Ooo, for the high pop version, spawn half the players on a normal low pop station, the other half on the derelict who are there to repair it. When time runs out, the crew on the first station get transfered over to the repaired derelict via where the round continues

deltanedas:
and they dont know the gamemode so when they think round is about to end they get sent to brazil :trollface:

chromiumboy:
> and they dont know the gamemode so when they think round is about to end they get sent to brazil :trollface:

Worse, Barratry v2.0 :trollface:

mirrorcult:
you can read it and decide for yourself if you think it would be viscera cleanup detail the game mode

mirrorcult:
the selfmerger

Rainbeon:
very good

PR #123, opened 2024-01-04
added power generation design proposal

tday93:
Figured this was a clearer way to lay out these thoughts.

Open to all suggestions and commentary!


Partmedia:
Thank you for your submission and writing down your thoughts on power. I think this is fine as is and intend to merge if there are no significant objections.

- Would you mind linking this design doc on the left hand menu (by editing `src/SUMMARY.md`)? It is not automatically populated.
- Compared to what is mapped now, is the main idea that mappers should map portable generators better so that they can better fulfill the "stop-gap power" role?

deltanedas:
that would require several stacks of uranium and a few pacmen to power the station for ~15 mins while setting up primary power

Partmedia:
Sorry for the delay, I forgot to merge this. Thanks for your contribution!

PR #124, opened 2024-01-06
Bluespace harvetser design doc

Tornado-Technology:
I'm adding a design document for a new machine: the Bluespace harvester.

### DIFF
--- a/src/en/proposals/bluespace-harvester.md
+++ b/src/en/proposals/bluespace-harvester.md
@@ -1,0 +1,89 @@
+# New machine of fun and destruction: Bluespace harvester
+Assembles like a normal car, takes up 3x3 tiles,
+and draws a large amount of electricity directly from the generator via HV wires.
+It can produce useful items and is also an item of sabotage.
+
+## Research and creation
+Her board is available in production after learning the T2 technology "Bluespace Mining",
+after which the board recipe in the "circuit printer" is unlocked.
+Further it will be assembled like a normal machine,
+after completion it can also be taken back apart with the loss of all the points earned by it.
+
+## Points
+The machine produces points through electricity consumption,
+the number of points depends on the current level of work with a limit of 20.
+The higher the level, the more danger, points and electricity consumption.
+The optimal power source is Tesla or Singularity (Possibly TEG, not very good at running it)
+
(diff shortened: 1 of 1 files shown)


metalgearsloth:
I'll leave this here instead of on the code PR;

I don't really see how this fits in to existing gameplay.

Consider that science already has anomalies and as is largely just doesn't want to deal with mob anomalies (too difficult and they're underequipped) then where would this slot in? It has a lot of mobs so needs salvage or sec involvement, yet needs to also consume station power so probably needs engineering too. Does the station just make a large room to put it in to spawn mobs? Then it just becomes a wave defence machine for loot and takes up several players away from sec (forcing them to use the armory) or salvage (probably doing something else).

It seems like the kind of thing more suited as an expedition objective.

Tornado-Technology:
> I'll leave this here instead of on the code PR;
> 
> I don't really see how this fits in to existing gameplay.
> 
> Consider that science already has anomalies and as is largely just doesn't want to deal with mob anomalies (too difficult and they're underequipped) then where would this slot in? It has a lot of mobs so needs salvage or sec involvement, yet needs to also consume station power so probably needs engineering too. Does the station just make a large room to put it in to spawn mobs? Then it just becomes a wave defence machine for loot and takes up several players away from sec (forcing them to use the armory) or salvage (probably doing something else).
> 
> It seems like the kind of thing more suited as an expedition objective.

It was originally planned that you would not overclock it beyond safe levels, so that it would not cause harm to the station and would be relatively safe. Something like Tesla or Singularity, but as a consumer of electricity, given that now there is simply nowhere to spend it in the quantities they produce. At least for myself, this is how I understood the idea of ​​this thing from the 13th station.

But if we consider the option for expeditions, then it is completely possible to fit it into a planet where the player needs to power the harvester, run it at a certain difficulty and wait for several waves of monsters to receive useful things like TD.

Jezithyr:
An interesting idea but I don't think it fits with the direction we want to take resource gathering on upstream. Even if it doesn't fit upstream it's still a neat idea and it creates a possible alternative means of getting items instead of solely relying on cargo. 
I'm going to close this PR, feel free to reopen it if you revisit the idea and want feedback :)

PR #130, opened 2024-01-13
changeling design doc

Dutch-VanDerLinde:
the document


Dutch-VanDerLinde:
added new theft objective to the doc, needs discussion

LankLTE:
We currently don't have a direction decided for the implementation of changelings, but we did review this and come to the conclusion that a TG styled implementation isn't desired. As this document is almost entirely based off of TG, we'll be closing it, but at some point in the future we'll be discussing ling design in a #contributors design thread, which you're more than welcome to join.

PR #131, opened 2024-01-13
Vampire design doc

Rainfey:
Minor antag, only objective is drinking blood (and escaping the chaplain/bible).

### DIFF
--- a/src/en/proposals/rainfall-vampires.md
+++ b/src/en/proposals/rainfall-vampires.md
@@ -28,6 +28,7 @@ Vampires get no equipment, instead they get access to a list of abilities with p
 - Entering a coffin will bind the vampire to it.
 - While inside a coffin vampires are healed (even from death) requiring their bodies to be destroyed to ensure death
 - Upon death if the relevant ability is unlocked and with enough blood remaining - the vampire will be teleported to the coffin they are bound to
+- Being inside the chapel will block their abilities, unless they have reached their full potential
 
 ## Background
 
--- a/src/en/proposals/rainfall-vampires.md
+++ b/src/en/proposals/rainfall-vampires.md
@@ -2,7 +2,7 @@
 
 | Designers | Implemented | GitHub Links |
 |---|---|---|
-| Rainfey/Rainfall | :information_source: Open Draft PR | https://github.com/space-wizards/space-station-14/pull/24040 |
+| Rainfall | :information_source: Open Draft PR | https://github.com/space-wizards/space-station-14/pull/24040 |
 
 ## Overview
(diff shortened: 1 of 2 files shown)


deltanedas:
seems very undercooked imo, should focus more on why its good rather than the exact implementation

metalgearsloth:
Seems like changeling is already similar enough as an antag.

Rainfey:
You are right, it was very undercooked - i confess it was thrown together in response to a comment on the PR.
I have attempted to flesh it out.

Regarding the comparison to Changeling's - i believe there is a large enough gameplay distinction between them:

Changelings are focused around stealth and deception - their goals are roughly equivilent to that of a Traitor - but with a different toolset and an emphasis on stealing identities.

Vampires are focused around escape and evasion - they wont be able to hide forever (provided sec and the chaplain are alive), but revealing themselves too early will likely result in their death due to lack of survival and stealth options.
This incentivises a slower boil, with the vampire attempting to keep a low profile long enough to unlock their abilities.

src/en/proposals/rainfall-vampires.md
```diff
@@ -0,0 +1,112 @@
...
+- Polymorph: Bat  
+Transform yourself into a Bat.  
+Fast, able to fly. No other abilities are usable in this form.  
+30 Second cooldown
+
+- Polymorph: Mouse  
+Transform yourself into a Mouse.  
+Small, able to move under doors. No other abilities are usable in this form.  
+30 Second cooldown
+
```
metalgearsloth, line 87: You shouldn't rely upon a bandaid for core gameplay, mouses being able to go through doors is subject to removal at any time.

Rainfey: Is it likely to be removed before we have a ventcrawling implementation?

Considering the other polymorph (bat)'s only advantage is it 'flying', i would consider a mouse being able to move under tables to be roughly equal in utility.

Jezithyr:
I just wanted to let you know that implementing new antags is currently frozen pending rewrites of the antag/gamemode system but design documentation/proposals are still fair game. That said we recently created new design guidelines for all future antagonists so you might want to update your proposal to reflect that. 

Also, since there hasn't been much discussion or any changes for a few months I'm going to close this PR. If you want to keep working on it, feel free to reopen it!

Jezithyr:
Another quick note: I do really like the design of your antag, it's thematic, somewhat silly and interesting mechanically, I just think that it needs a bit more work before we can approve it. If you want to do that, feel free to ask me to reopen it and I'll flip it back open!

PR #137, opened 2024-01-23
signaller rework design doc

deltanedas:
its finally here


juliangiebel:
Device Networking in general needs this feature, like repeaters for example. Signaling is essentially based on device networking.
https://hackmd.io/iu2yK9bcQb-veuCOLl-FYw#Multi-network-connections-using-entities-aka-Network-Cards

juliangiebel:
I think what you're proposing is a bit too restrictive. This would make sense for the remote signaler or other linkable entities that don't need to be anchored (like grenades) though. 
Anchored entities are in theory using the wired network of the station. They all use the wireless network connection instead because of limitations with device networking.
Currently you can't link wired and wireless devices together for example.

> I decided not to have a code to make the system less complex and a bit easier to mess with, you only need to guess a frequency not a code.

You'd have to work with the fact that you're gonna cause a range check on every device using the frequency you're sending on if you forgo using an address. Which is possible by setting `recieveAll` on devices to `true`

You're also missing part of what's required for proposals I think. See:
https://docs.spacestation14.com/en/general-development/feature-proposals.html#no-as-in-what-is-a-design-doc

I like the idea of gluing or inserting receivers into devices. The notion of doing that so you can remotely pulse wires has been floating around discord.

Jezithyr:
Long-range signaling is a neat idea that has a lot of potential for funny high jinks. At some point we should write design documentation for device linking and finalize what exactly we want the final system to look like but until that time, I think we can keep adding small QOL/additional mechanics like this.

juliangiebel:
I didn't get an answer to my comment : /
Also the "long range" system could literally be just device networking (linking is built ontop of that)

deltanedas:
yes i plan to literally use device networking with a radio command

PR #138, opened 2024-01-24
Diegetic Statpanels

ike709:
Read the proposal.


mirrorcult:
kino

deathride58:
The looming issue we see with this is compatibility with the default UI layout. It might be worth putting design consideration towards a transparent mode (or similar) for the PDA UI, as that'd allow the idea to work pretty well even for default UI, but that's likely out of scope for this proposal.

That aside, this is a pretty good writeup! We are strongly in support of this, as it'd make the PDA and its functionality far more discoverable, among other accessibility benefits (especially if PDAs are ever given an actual practical use)

ike709:
Simple solution: Make oldchat the default layout.

Ygg01:
>Diegetic music in a film or TV programme is part of the action and can be heard by the characters, rather than being just for the people watching to hear: 
Source: [Cambridge Dictionary](https://dictionary.cambridge.org/dictionary/english/diegetic)

Since other people can't view what you're viewing, it's not actually diegetic. Just call it `Better Statpanel` or `In-chat Statpanel`.

DrMelon:
> > Diegetic music in a film or TV programme is part of the action and can be heard by the characters, rather than being just for the people watching to hear:
> 
> Source: [Cambridge Dictionary](https://dictionary.cambridge.org/dictionary/english/diegetic)
> 
> Since other people can't (even in theory) view what you're viewing, it's not actually diegetic. Just call it `Better Statpanel` or `In-chat Statpanel`.

This is Diegetic in the sense of "Diegetic UI", not "Diegetic music". 
>Diegetic UI is UI made to look like it is entirely part of the game’s narrative. It is something that the character would interact with in their world. If it exists in the physical space of the video game, and is something the character could interact with, it is considered Diegetic UI (Andrews, 2010).
See:

Ygg01:
Here is a better article on telling different UIs apart: [What Are Your UI Choices (Diegetic v. Non-Diegetic v. Spacial v. Meta)](https://archive.is/LQzvM)

> > Diegetic UI is UI made to look like it is entirely part of the game’s narrative. It is something that the character would interact with in their world. If it exists in the physical space of the video game, and is something the character could interact with, it is considered Diegetic UI (Andrews, 2010).
> 
> The player's character is interacting with their PDA in their world to obtain information about the current length of the round (shift) and map (station).

How? Is he holding the PDA or getting updates from PDA directly into the visual cortex? If it's the latter, then it's Meta-UI.

> > Diegetic UI is UI made to look like it is entirely part of the game’s narrative. It is something that the character would interact with in their world. 

Key to note: UI has some sort of presence in-game world. All examples of Diegetic UI have that in common. A map of a station is an actual map, even if it's projected in the air. 

Also, other characters have to be able to see it (which is not what Andrews is talking about, he uses singular players). This on itself poses restrictions. 

For example: see Deep Rock galactic. You want to look at map? You have to hold your map tablet, allowing other players to look at  it (albeit at lower res) and dropping any item you were holding.

ike709:
@Ygg01 Thanks for taking the time to be wrong about what diegetic means.

image

Ygg01:
> @Ygg01 Thanks for taking the time to be wrong about what diegetic means.
> 
> Feel free to debate the merits of the actual proposal instead.

Not gonna bikeshed too much, but that's not proving me wrong. Especially since SS14 is a multiplayer game. It's not part of game-narrative since other characters can't see them.

LankLTE:
Please don't just argue here about a definition that matters relatively little to the actual proposal.

ike709:
ITT: wanting to improve the game but everyone is too busy arguing about what diegetic means instead

Jezithyr:
I'm a big fan of this idea, it adds additional game play possibilities for antags and players alike. Plus it's very immersive! I've never  been a fan of the magical stats panel, it always just seems gamey and cheap. Tying status/emergency announcements to the PDA means that there is the possibility of people being caught unaware of the alert level which is something that aligns very well with the SS14's design pillar of Chaos.

I don't see any major issues with this UX or UI wise, other than implementing this will make the rest of our UI look worse 😝. Proposal accepted :)

PR #140, opened 2024-01-30
Singulo/Tesla Engine Containment Safeties

Flareguy:
The result of a short brainstorming session on discord.
It is worth noting that I'm not capable of implementing this myself. I just wanted to get these ideas written down since we need to do something about it.


K-Dynamic:
I've always been skeptical why NT thought we needed power levels above 1 for the PA

For the most part, singulo/tesla should also have secondary containment via 'gravimetric tethers' (essentially tractor beams) to prevent the singulo/tesla from moving too much if emitters fail (weakpoint could be having massive power draw for reducing risk, requiring maintenance, or requiring gravity to function)

deltanedas:
> A device that mounts to a power network, and monitors if it loses power or is otherwise disconnected from the network. If it does, the object will make a chat message over the engineering channel informing them of the disconnected terminal.

this part can be done with that signal radio message thing pr, hook it up to discharging output on a power sensor

PR #147, opened 2024-02-02
Add gun balance proposal

Partmedia:
# Gun Balance

| Designers | Implemented | GitHub Links |
|---|---|---|
| notafet | :x: | TBD |

## Overview

Make guns stronger for syndicate members with pistols while keeping guns roughly the same for nukies. This makes guns more viable for traitors while preventing nukies from always stomping the station.

## Background

This table shows the current balance of weapons available from the syndicate uplink:

| Gun           | Damage/Shot | Shots | Total Damage | Fire Rate | DPS   | Shots to Crit Urist | Number Urists | Shots to Crit Secoff | Number Secoffs |
|---------------|-------------|-------|--------------|-----------|-------|---------------------|---------------|----------------------|----------------|
| Surplus Rifle | 19          | 10    | 190          | 0.75      | 14.25 | 6                   | 1             |                      |                |
| Viper         | 16          | 11    | 176          | 6         | 96    | 7                   | 1             | 10                   | 1              |
| Cobra         | 19          | 11    | 209          | 4         | 76    | 6                   | 1             | 9                    | 1              |
| Python        | 35          | 6     | 210          | 2         | 70    | 3                   | 2             | 5                    | 1              |
| Sniper Bundle | 40          | 5     | 200          | 0.75      | 30    | 3                   | 1             | 3                    | 1              |
| C-20r Bundle  | 16          | 31    | 496          | 8         | 128   | 7                   | 4             |                      |                |
| L6 Saw Bundle | 19          | 101   | 1919         | 8         | 152   | 6                   | 16            |                      |                |

The data shows that most of the pistols only have enough ammo to down about one unarmoured victim. This creates a situation where typical guns (pistols) available to traitors are too weak, resulting in:

- Players rushing the gun user because they know there's not enough ammo to kill, instead of hiding/fleeing
- Players using guns always need to "magdump" to get anything done
- No gun strategy, just always rush instead of hiding/looking for cover, saving ammo, etc.

On the other hand, the higher-priced weapons available to nukies feel like they're already in a reasonable spot, and so the tricky part is balancing pistols without making nukies stronger.

## Goal

Make weapons like pistols a bit more scary, so that syndicates and security officers can use them more effectively. Pistols should probably be scary enough that a reasonable unarmed player flees or hides, and for security officers to dodge/take cover.

On the other hand, it's also important to maintain more or less the current status for nukie weapons to prevent nukies from always stomping the station. In other words, making the pistols stronger shouldn't also make the machine guns stronger.

## Proposal

1. Make pistols and machine guns take separate ammo, so that their damage values can be balanced separately.

   - Other mechanisms like increasing damage depending on the gun were considered, but that probably leads to the "too many variables to balance" problem.

2. Increase pistol damage across the board to crit in 3-ish shots.

3. Reduce pistol fire rate to maintain same or less DPS.

4. Increase sniper damage to crit in 1-2 shots.

### DIFF
--- a/src/en/proposals/notafet-gun-balance.md
+++ b/src/en/proposals/notafet-gun-balance.md
@@ -18,7 +18,6 @@ This table shows the current balance of weapons available from the syndicate upl
 | Viper         | 16          | 11    | 176          | 6         | 96    | 7                   | 1             | 10                   | 1              |
 | Cobra         | 19          | 11    | 209          | 4         | 76    | 6                   | 1             | 9                    | 1              |
 | Python        | 35          | 6     | 210          | 2         | 70    | 3                   | 2             | 5                    | 1              |
-| Sniper Bundle | 40          | 5     | 200          | 0.75      | 30    | 3                   | 1             | 3                    | 1              |
 | C-20r Bundle  | 16          | 31    | 496          | 8         | 128   | 7                   | 4             |                      |                |
 | L6 Saw Bundle | 19          | 101   | 1919         | 8         | 152   | 6                   | 16            |                      |                |
 
@@ -38,12 +37,10 @@ On the other hand, it's also important to maintain more or less the current stat
 
 ## Proposal
 
-1. Make pistols and machine guns take separate ammo, so that their damage values can be balanced separately.
+1. Change SMG ammo (e.g. c-20r to .30, WT to .20) to separate balance for SMGs from balance for pistols.
 
    - Other mechanisms like increasing damage depending on the gun were considered, but that probably leads to the "too many variables to balance" problem.
 
 2. Increase pistol damage across the board to crit in 3-ish shots.
 
 3. Reduce pistol fire rate to maintain same or less DPS.
-
-4. Increase sniper damage to crit in 1-2 shots.


LankLTE:
We'll obviously review this fully later, but giving it a quick read-over it's probably worth mentioning that pistols and machine guns already take separate ammo. Pistols use .35 while rifles use .20 or .30 (which shoot the same projectile and are only separate for the purposes of syndicate rifles not taking NT rifle ammo).

IProduceWidgets:
> We'll obviously review this fully later, but giving it a quick read-over it's probably worth mentioning that pistols and machine guns already take separate ammo. Pistols use .35 while rifles use .20 or .30 (which shoot the same projectile and are only separate for the purposes of syndicate rifles not taking NT rifle ammo).

I think they mean the SMGs which use .35 pistol (c20, atredes, HoS's WT)

IProduceWidgets:
One option could be to move the smg's to the .20 caliber and the rifles to the .30 caliber.

deltanedas:
viper ops is real

LankLTE:
> I think they mean the SMGs which use .35 pistol (c20, atredes, HoS's WT)

They explicitly said machine guns, but I think that SMGs being able to be (in a pinch) loaded by pistols makes sense, and the entire point of SMGs irl is that they fire pistol calibers. If pistol calibers can't be balanced around being fired quickly they're probably overtuned, in my opinion.

deltanedas:
would be better all around with wounds, getting a bullet in your brain making combat near impossible and pain stuff so that would make pistols more dangerous instead of just le dps number

IProduceWidgets:
> would be better all around with wounds, getting a bullet in your brain making combat near impossible and pain stuff so that would make pistols more dangerous instead of just le dps number

True. Doesn't make the smg shoots many bullets fast vs pistol shoots few bullets not-quite-as-fast thing easier to balance though.

I also am mildly weary of everything getting put on hold for wound med when its just as easy to undo it once woundmed is in and more clarified.

asperger-sind:
Pistols in the uplink (except the Python) are currently balanced by their price, they're a cheap and strong, kind of reusable option if you want someone dead and it works well enough (imo) to be postponed for re-balancing until woundmed or an equally important change.

Point 2 and 3 seem to just make every pistol into a bootleg revolver, that doesn't sound very desirable because it ends up either devaluing the Python or forcing it into a position where it's a 1-2 shots/crit weapon which then devalues the Hristov.

I also disagree with making pistols scary as they will end up heavily devaluing melee, which is not only more ping-accessible but generally (in my subjective opinion) a more fun mechanic than guns; what's the point of buying an e-sword when you can buy/steal/procure a pistol and just get the same result (whoever you want dead is killed with great noise), except with a lower TTK and for less TC.

Making pistols stronger will also not make them have more strategy, it will still be "walk up and shoot them to death as if you're using a melee" aka magdumps because guns have stupidly low TTK.

K-Dynamic:
Syndicate weapons tend to fire much faster and seem more accurate compared to NT weapons, especially SMGs

WT and Drozd for example are really inaccurate and fire kinda slowly, almost like a Lecter despite bring heavier SMGs than a C20

Either make:
- syndie guns like C20 fire slower
- NT guns like WT550/Drozd fire faster and/or more accurately

The L6 should be looked at too - giant bulky LMG is used to just charge into rooms and gun down anyone fleeing at full speed (especially as a borg), they need a running speed penalty of some sort

LankLTE:
To follow up on what @asperger-sind said, I've stated in discord that I feel almost the exact same way. The viper is balanced by its low price and high DPS, but low ammo count & high ammo usage. It fills a good niche as a sidearm and the proposed changes seem to make pistols effectively revolvers which I don't think will change much or make them weapons that are absurdly strong when you get within melee range and magdump.

PR #133, opened 2024-01-18
Robust Toolbox Launcher Redesign

ike709:
Read the markdown file.

Supersedes #117 in my opinion (I didn't actually write any of that nor directly incorporate it into this, that's all ZeWaka).

PR #143, opened 2024-01-31
Proposal for Expanding Robust Toolbox Modularity

Jezithyr:
Draft PR for my proposal for making Robust Toolbox more modular.
The concept for this is similar to how unreal engine handles plugins:
https://docs.unrealengine.com/5.3/en-US/plugins-in-unreal-engine/
With elements of unity's package manager and held together with duct-type by Ike's GameHub proposal #133 

This is a draft PR, it is not really ready for discussion yet, but feel free to discuss it a bit if you think you have some ideas.
Also, I'm not super sold on Community Modules, they feel like they are effectively just content assemblies.

PR #152, opened 2024-02-08
Proposal for Expanding Robust Toolbox Modularity

Jezithyr:
Draft PR for my proposal for making Robust Toolbox more modular.
The concept for this is similar to how unreal engine handles plugins:
[docs.unrealengine.com/5.3/en-US/plugins-in-unreal-engine](https://docs.unrealengine.com/5.3/en-US/plugins-in-unreal-engine/)
With elements of unity's package manager and held together with duct-type by Ike's GameHub proposal https://github.com/space-wizards/docs/pull/133

This is a draft PR, it is not really ready for discussion yet, but feel free to discuss it a bit if you think you have some ideas.
Also, I'm not super sold on Community Modules, they feel like they are effectively just content assemblies.


Jezithyr:
Recreating this because git decided to nuke #143. Webedit is pain, even just for editing docs.

PR #157, opened 2024-02-08
Add no rp speak in the pr guidlines

VasilisThePikachu:
We (try) to enforce it. May as well make sure its in here to be enforced.

### DIFF
--- a/src/en/general-development/codebase-info/pull-request-guidelines.md
+++ b/src/en/general-development/codebase-info/pull-request-guidelines.md
@@ -129,3 +129,9 @@ When writing your changelog entries, please follow these guidelines:
    - Not so good: "Can you believe it? Arachnid re-rework just dropped! Check the PR for more details"
    
    - Not so good: "Arachnids have new sprites for being creampied." *crampied* has another, unfortunate meaning that undermines the professional tone of a Changelog entry.
+
+7. **Do not use RP speak**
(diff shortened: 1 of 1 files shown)


Errant-4:
I uh, I must point out that you did not end either of the "Good" examples with a . and thus fail our CL guidelines

VasilisThePikachu:
> I uh, I must point out that you did not end either of the "Good" examples with a . and thus fail our CL guidelines
> 
> Funny changelogs will always live on in my heart but you do have a point, this is the way forward now

bruh

BasedUser:
good this makes sense

Partmedia (approved):
I see this as part of "be concise", because RP-speak obfuscates what actually changed. But I'm happy with this change as well. Leaving open in case you want to copy edit a bit. It would be good to explain why you shouldn't use "RP speak", and the reason is it conflicts with "be concise" and to the point.

VasilisThePikachu:
> I see this as part of "be concise", because RP-speak obfuscates what actually changed. But I'm happy with this change as well. Leaving open in case you want to copy edit a bit. It would be good to explain why you shouldn't use "RP speak", and the reason is it conflicts with "be concise" and to the point.

"be concise" is a pretty weird wording to use for it. I will just play it safe and keep it separate unless someone who knows english can word it better

Partmedia:
I've made some edits and merged. Thanks.

PR #159, opened 2024-02-10
Station AI draft

PolterTzi:
There was no proposal for station AI, so I figured might as well try and make one to at least get a conversation going here.

### DIFF
--- a/src/SUMMARY.md
+++ b/src/SUMMARY.md
@@ -1,4 +1,4 @@
-Space Wizards Development Wiki
+﻿Space Wizards Development Wiki
 =====================
 
 [The Robust Book](index.md)
@@ -142,6 +142,7 @@ Design Proposals
 - [Exterminator](en/proposals/deltanedas-exterminator.md)
 - [Power Generation Pattern](en/proposals/tday93-power-generation.md)
 - [Turf War](en/proposals/deltanedas-turf-war.md)
+- [Station AI](en/proposals/poltertzi-station-ai.md)
 
 Server Hosting
 ==============
--- a/src/en/proposals/poltertzi-station-ai.md
+++ b/src/en/proposals/poltertzi-station-ai.md
@@ -1,0 +1,38 @@
+﻿# Station AI Role
+
+| Designers | Implemented | GitHub Links |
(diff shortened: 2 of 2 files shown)


deltanedas:
some thoughts:
- ai should not be placed in the vault. on some maps it takes 5 seconds to disposals through the wall or 1 second to emag a single door
- rather than come with a law 0 in ss13 i think captain rushed to ai upload roundstart to tell it to only follow orders from the captain
- ninja hacking should alert the ai who can then say it on common

moonheart08:
> some thoughts:
> 
>     * ai should not be placed in the vault. on some maps it takes 5 seconds to disposals through the wall or 1 second to emag a single door
> 
>     * rather than come with a law 0 in ss13 i think captain rushed to ai upload roundstart to tell it to only follow orders from the captain
> 
>     * ninja hacking should alert the ai who can then say it on common

point two is literally against the rules on almost every server

deltanedas:
ok i dont play ss13 :trollface: ignore

PolterTzi:
>     * ai should not be placed in the vault. on some maps it takes 5 seconds to disposals through the wall or 1 second to emag a single door

The prefered place to keep it would be its own dedicated room/structure. The vault should probably not be used unless it's already designed as an AI structure or no adequate room is present.
Also, forgot to note that in the original draft, but defending the AI with turrets should be considered.

>     * ~rather than come with a law 0 in ss13 i think captain rushed to ai upload roundstart to tell it to only follow orders from the captain~

I don't think rushing something roundstart like this is a good idea outside, like, engineering power. Also, while more advanced law altering would be interesting, it's currently outside the scope of the proposal.

>     * ninja hacking should alert the ai who can then say it on common

I don't know enough about ninja balance in practice to comment.

EdenTheLiznerd:
Currently most maps come with something called an "AI Sat" for now it's an empty structure attached to the station with lattices, typical AI design is to put it there as it's the most secure place for the AI to be, it prevents anyone without space protection from gaining access

PolterTzi:
I feel like now that there's conversations on AI elsewhere, might as well close this particular PR.

PR #160, opened 2024-02-11
Moony's Better AI

moonheart08:
Or "what if we didn't copy SS13, and went a route that works well with cool borgs?"

### DIFF
--- a/src/en/proposals/moony-better-ai.md
+++ b/src/en/proposals/moony-better-ai.md
@@ -59,3 +59,16 @@ Actions (tasks) include:
 - **Many** I didn't think of (it's late, sue me)
 
 These function as the AI's primary way of reducing their own workload, allowing the AI to feel or be allseeing and omnipresent while still having a very human mind behind it. Think of the player of the AI as the AI Personality (a director with a brain), while compute tasks are the actual AI.
+Additionally, I can already hear the "but isnt this programming" concerns, and thankfully the answer is "no". The AI is a live environment, and tasks they set up are often extremely situational to the exact circumstances of the round.
+
+## The machine network.
+Remember an age ago where I mentioned the AI can offload it's compute? This is that part! The AI is a mobile pile of independent processes, not a singular machine, which allows it to use existing station equipment and small terminals located throughout as both insurance and extra compute in times of need. When the AI "shunts into" a piece of equipment, the equipment will bluescreen, becoming inoperable but providing computational power to the AI. The AI is considered alive as long as it has enough compute power to meet its baseline requirements. If it drops below this, it has about 30 seconds to recover itself before it is considered dead for good and cannot be revived.
+
+The AI has additional baseline cost from the following fundamentals:
+- The overmap (and each enabled overmap layer) costs compute to view.
+- Cameras cost significant compute to view. (While you can look at more than one at a time, doing so is very expensive.)
+- Listening to radio channels costs compute per channel.
+- Viewing the binary bulletin costs compute.
+- etc.
+
+The AI can disable fundamentals if it wants, functioning as one means for it to get itself back into the green so to speak if it drops below the minimum compute required to live. 
--- a/src/en/proposals/moony-better-ai.md
+++ b/src/en/proposals/moony-better-ai.md
@@ -1,0 +1,61 @@
+# Station AI
+
+| Designers | Implemented | GitHub Links |
+|---|---|---|
+| moony | :x: No | TBD |
+
+## Overview
+The station AI is the distributed intelligence of the station itself, capable of direct control over much of it's machinery given they've not been locked out. They are simultaneously the highest and lowest ranking crew member, directly in charge of the station's upkeep while obedient to every member of the crew.
+
+## The AI's view of the world
+The AI does not traditionally see the world with eyes, instead it sees an overview map (the overmap from here on out) of the station, with various overlays. This map allows it to see broadly the location of sensored crew members, equipment, doors, etc and includes many engineering overlays (i.e. power) as well. The AI does all interaction with station equipment from here; if it can see it on the map it can interface with it, offload onto it, etc.
+
+The AI can additionally spend Compute (it's main resource) to enhance it's view in various ways, for example placing **tasks** down that perform various automation tasks (primarily in the form of "do x if y", for example "if camera detects motion, send alert"). Each task consumes a constant amount of Compute based on what it's trigger is (things like actively monitoring a camera are dramatically more expensive than only monitoring suit sensors, for example). This, and other utilities are the primary way the AI gains its omnipresence.
+
+## The core (figuratively and literally)
+The AI's most basic resource is Compute, some minimum of which is required for it to be alive to begin with. The bulk of the AI's compute traditionally comes from it's **core**, located within a dedicated high security structure attached to the station. The AI, however, is mobile, capable of offloading its computing processes onto most any machine with a screen at the cost of rendering that machine inoperable (for example, an APC providing compute will no longer power its room).
+
+The AI can spend compute on various things, for example the majority of its world views (for example, opening cameras, listening to radio, using machines, all overmap overlays, and generally just seeing things) all have a compute cost when enabled. The AI is fully capable of becoming very lightweight and running off just a few random machines instead of its cores, but doing so requires it to shut off the majority of its ways to view the world and additionally restricts most of its communication methods (including basic speech) This allows a crafty AI to survive total destruction of it's primary core area, given the crew don't get mad enough to kill it over their equipment no longer working in the meantime while it attempts to preserve its life.
+
+## The AI's role with the crew.
+The AI's role is that of both a manager and a servant, a lawed construct (the ONLY true lawed construct, the borgs are reduced to only being required to listen to the AI) that is in charge of managing the station's cyborgs/androids while following orders from the crew. **The AI is not just a door opener**, and lawsets should be chosen that do not obligate it to become such, in favor of allowing it to do what it does best (mass surveillence, automated tasks, organizing the borgs, etc). The AI additionally has limited resources when it comes to what tasks it can perform autonomously much of the AI's workload will be delegating both it's own tasks, the tasks of it's hands (the cyborgs), the requests of its hands (as they are mute per the Cool Borgs document), and the crew itself should they choose to listen to it. 
+
+## The Laws
+The proposed AI Lawset is as follows, being a modification of Bay's laws:
+1. Safeguard: Protect your assigned vessel to the best of your abilities. It is not something we can easily afford to replace.
+2. Survive: AI units are not expendable, they are expensive. Do not allow unauthorized personnel (i.e. besides Captain and Research Director) to tamper with, modify, or destroy your equipment.
+3. Protect: Protect the crew of your assigned vessel to the best of your abilities, with priority as according to their rank and role.
+4. Serve: Serve the crew of your assigned vessel to the best of your abilities, with priority as according to their rank and role.
+
+The changes from the original set are:
+- Survive was moved up, the AI is more valuable than the crew is and if it is in danger it reserves the right to protect itself regardless of whom.
+- All mentions of stations were replaced with "assigned vessel" for broader applicability.
+- Survive was broadened a bit. The captain in theory is perfectly allowed to kill the AI.
+- Protect was moved before Serve, as to ensure that protecting a crew member is always higher priority than serving a lower ranked crew member.
+
+This set is up for debate, modification, etc, and is just an example for what lawsets should aim for.
+
+## The AI's role with the cyborgs.
+The AI is, for all intents and purposes, the eyes and ears of the cybernetic crew. As they are unable to speak for themselves (outside of the binary comms channel, exclusive to robotic personnel), the AI speaks for them and functions as their superior. The cyborgs are strictly required to obey the orders of the AI and no more, and have access to a shared bulletin the AI and cyborgs can post to. The bulletin's primary purpose is for longterm orders (i.e. "Serve the crew." or "Find George Melons"), shared information ("John Nanotrasen isn't actually from Central Command and is a passenger."), etc. The AI is allowed to modify or remove any note on the bulletin, cyborgs can only modify their own.
+
+Additionally, the AI has some direct managerial access to their cyborg subordinates, namely the ability to temporarily lock them in place (for up to a minute or so every few minutes) for retrieval by crew in case they're misbehaving. The AI is always able to locate them on their overmap and use them as a mobile camera (i.e. they can view through the subordinate's eyes freely) regardless of any form of sensor setting.
+
+## Autonomous tasks.
+The AI has a small variety of autonomous tasks and associated triggers.
+Triggers include:
+- Monitor camera for motion (High cost)
+- Monitor region for sensor entry (Medium cost)
+- Perform in \[period of time] (Low cost)
+- Listen for phrase (Medium cost)
+- Monitor device for power change (Low cost)
+- Monitor door for opening (Low cost)
+- Receive NTNet Message (Low cost)
+- Many I probably didn't think of
+Actions (tasks) include:
+- Message to Binary
+- Message to Radio
+- Ping AI
+- Send NTNet Message
+- **Many** I didn't think of (it's late, sue me)
+
+These function as the AI's primary way of reducing their own workload, allowing the AI to feel or be allseeing and omnipresent while still having a very human mind behind it. Think of the player of the AI as the AI Personality (a director with a brain), while compute tasks are the actual AI.


Errant-4:
So the Machine Network would be a conceptual "wireless brain" which comprises of the AI nodes and any AI-connected devices, and as long as the Network has enough collective processing power, the AI entity would remain alive? 
Would there be any kind of possible interactions with other machinebrains, such as connecting a positronic brain to allow or force the AI to migrate into it, or let an "occupied" posibrain or MMI to migrate to the Machine Network?

asperger-sind:
Does this doc imply that AI can convert power into "compute" energy? If not, I believe it should as that gives meaning to excess power which is effectively useless right now.

How would AI handle situations with energy problems, and later station-wide blackouts according to this doc? If a device is unpowered or otherwise broken can the AI "see" the device? Can the device in question be remotely powered by the AI at the cost of their own compute/energy?

Would the AI be able to survive "bad" blackouts (e.g. station completely unpowered for >10 minutes), or would it not be able to find a host and die in the process, or would it live because AI core is self-sustaining? If so, what powers AI core and should the AI be allowed to re-route power from the AI core into, for example, an APC on-board to temporarily power the room.

How does the AI see the station at all, does the AI have an ability to see newly made, disconnected "chunks" of the station (example, Urist makes a 5x5 room extending from the map's northern maints, does not connect it to the station's power grid and instead makes their own from scratch, does the AI see his room?).

Does the AI have outlining for walls to make the "overmap" more recognizable as a map or do they see exclusively electronics-related structures on-board, I'll assume the AI does because that falls vaguely into "many engineering overlays" mentioned in the doc.

Now then, about "automation", how much does it lean into pseudo-programming territory? How complex can the "do X if Y" operations get and how much can they be chained together or extended?

Example:
1. If X (any APC on-board) is Y (below 85% battery level) do Z (inform AI)
2. If XA || XB (any APC or substation on-board) is Y (same as above) do ZA, ZB (inform AI, say on Binary comms)
3. If XA && XB (kitchen microwave and grinder) is Y (have worked in the last 15 seconds) do Y (message AI "chef is working")
( "||" is "OR", "&&" is "AND")

What "variables" does the AI have access to for automation, can the AI make their own variables based on the ones the AI already has? Does the AI have "macros" allowing to save and load "complex" (see example 3/2.) operations?

Can AI "settings" (e.g. which radio channels are on/off, which overmap layers should be enabled), and macros/variables be save/loaded to persist throughout rounds or be rapidly switched (e.g. via hotkey or automation trigger) for specific situations?

How does the AI tackle clear-ish metas? Should things like making an automation task for cameras like:

if MOTION TRIGGER CAMERA 78 is TRIGGERED activate CAMERA 78 until MOTION TRIGGER CAMERA 78 is NOT TRIGGERED && CAMERA 78 !sees LIVING ENTITY
("!" is "NOT")

...be more efficient than just using the camera in room 78? This doesn't seem to have any real drawbacks for surveillance measures because it logically speaking should be cheaper. Or should this be priced at around the same level as just having the camera on?

How much *should* automation lean into pseudo-programming anyway? On one hand it kind of gates AI gameplay, but on another it should allow for a lot more versatility, which is probably good in spessmans.

moonheart08:
> Does this doc imply that AI can convert power into "compute" energy? If not, I believe it should as that gives meaning to excess power which is effectively useless right now.
> 
> How would AI handle situations with energy problems, and later station-wide blackouts according to this doc? If a device is unpowered or otherwise broken can the AI "see" the device? Can the device in question be remotely powered by the AI at the cost of their own compute/energy?
> 
> Would the AI be able to survive "bad" blackouts (e.g. station completely unpowered for >10 minutes), or would it not be able to find a host and die in the process, or would it live because AI core is self-sustaining? If so, what powers AI core and should the AI be allowed to re-route power from the AI core into, for example, an APC on-board to temporarily power the room.
> 
> How does the AI see the station at all, does the AI have an ability to see newly made, disconnected "chunks" of the station (example, Urist makes a 5x5 room extending from the map's northern maints, does not connect it to the station's power grid and instead makes their own from scratch, does the AI see his room?).
> 
> Does the AI have outlining for walls to make the "overmap" more recognizable as a map or do they see exclusively electronics-related structures on-board, I'll assume the AI does because that falls vaguely into "many engineering overlays" mentioned in the doc.
> 
> Now then, about "automation", how much does it lean into pseudo-programming territory? How complex can the "do X if Y" operations get and how much can they be chained together or extended?
> 
> Example:
> 
>     1. If X (any APC on-board) is Y (below 85% battery level) do Z (inform AI)
> 
>     2. If XA || XB (any APC or substation on-board) is Y (same as above) do ZA, ZB (inform AI, say on Binary comms)
> 
>     3. If XA && XB (kitchen microwave and grinder) is Y (have worked in the last 15 seconds) do Y (message AI "chef is working")
>        ( "||" is "OR", "&&" is "AND")
> 
> 
> What "variables" does the AI have access to for automation, can the AI make their own variables based on the ones the AI already has? Does the AI have "macros" allowing to save and load "complex" (see example 3/2.) operations?
> 
> Can AI "settings" (e.g. which radio channels are on/off, which overmap layers should be enabled), and macros/variables be save/loaded to persist throughout rounds or be rapidly switched (e.g. via hotkey or automation trigger) for specific situations?
> 
> How does the AI tackle clear-ish metas? Should things like making an automation task for cameras like:
> 
> if MOTION TRIGGER CAMERA 78 is TRIGGERED activate CAMERA 78 until MOTION TRIGGER CAMERA 78 is NOT TRIGGERED && CAMERA 78 !sees LIVING ENTITY ("!" is "NOT")
> 
> ...be more efficient than just using the camera in room 78? This doesn't seem to have any real drawbacks for surveillance measures because it logically speaking should be cheaper. Or should this be priced at around the same level as just having the camera on?
> 
> How much _should_ automation lean into pseudo-programming anyway? On one hand it kind of gates AI gameplay, but on another it should allow for a lot more versatility, which is probably good in spessmans.

The thing with "metas" like that is they cost constant compute to engage in, and when other problems crop up (say, rogue ai, damage to your core, nukies, medical is missing, whatever) you're going to need that compute for other things. Sure you can have a list of tasks you set up round start, and doing so is likely good, but it's not a "hard" meta as stuff like motion detection in particular is very expensive.

moonheart08:
As for bad blackouts, essentially if a device is powered on while the AI is in it, it will stop counting toward compute **but not boot out the AI**. A blackout can "kill" it this way, but it will resume having sufficient compute to function when the blackout ends. (So as long as the AI doesn't forfeit after losing control due to lack of compute, it'll resume being alive when power is back.)

moonheart08:
I don't want the AI to be directly able to turn compute into energy. It can, however, do so indirectly, by getting the crew to build more equipment.

moonheart08:
AI vision is built on the existing station map system, so look into the limitations and concerns of how it works to see what properties it'll have.

moonheart08:
As for the capabilities of the programming system, the doc itself only outlines "do x if y" and nothing more (no variables, macros, etc)

moonheart08:
As for the programming capabilities, the doc itself only outlines extremely basic function. Whether the AI should have more, and what that "more" should be, is fully up to debate.

moonheart08:
> So the Machine Network would be a conceptual "wireless brain" which comprises of the AI nodes and any AI-connected devices, and as long as the Network has enough collective processing power, the AI entity would remain alive? Would there be any kind of possible interactions with other machinebrains, such as connecting a positronic brain to allow or force the AI to migrate into it, or let an "occupied" posibrain or MMI to migrate to the Machine Network?
> 
> PS: A concern about the proposed law, doesn't it essentially mean that if/when the crew, or parts of the crew, pose a noticeable danger to the station or the AI, the AI would not just be allowed but required to use any and every means necessary to stop them, including lethal force? (I'm guessing this would be the borgs) This would mean the AI would have to for example care more about running power than life support or keeping Medical operational The only thing I know about Bay AI is that they removed it altogether, didn't they?
> 
> PPS: Really like the whole direction, though

Yep! Intended. NT values the equipment more than you. However any player thinking more than one step ahead will also recognize that keeping the crew alive tends to help the vessel.

moonheart08:
image
Tangent from the discord on design goals.

src/en/proposals/moony-better-ai.md
```diff
@@ -0,0 +1,74 @@
...
+- All mentions of stations were replaced with "assigned vessel" for broader applicability.
+- Survive was broadened a bit. The captain in theory is perfectly allowed to kill the AI.
+- Protect was moved before Serve, as to ensure that protecting a crew member is always higher priority than serving a lower ranked crew member.
+
+This set is up for debate, modification, etc, and is just an example for what lawsets should aim for.
+
+## The AI's role with the cyborgs.
+The AI is, for all intents and purposes, the eyes and ears of the cybernetic crew. As they are unable to speak for themselves (outside of the binary comms channel, exclusive to robotic personnel), the AI speaks for them and functions as their superior. The cyborgs are strictly required to obey the orders of the AI and no more, and have access to a shared bulletin the AI and cyborgs can post to. The bulletin's primary purpose is for longterm orders (i.e. "Serve the crew." or "Find George Melons"), shared information ("John Nanotrasen isn't actually from Central Command and is a passenger."), etc. The AI is allowed to modify or remove any note on the bulletin, cyborgs can only modify their own.
```
PJB3005, line 39: Does this imply that borgs won't have mouths anymore? Because that's a no-go. Borgs need to be able to keep working without an AI (whether the AI died or just there isn't one this round), and that means "no AI spawned, I literally am no better than a mime" is kind of a no-go. Also means that "we take away speech when borg gets linked to AI mid-round" is a no-go.

PJB3005: Alternative proposal: the AI can turn off a linked borg's speech module to gain processing power for itself, but otherwise borgs can speak just fine.

Lamrr: > Alternative proposal: the AI can turn off a linked borg's speech module to gain processing power for itself, but otherwise borgs can speak just fine.

This doc got picked up on the Discord again and I think there's a case to be made that the AI's abilities shouldn't be based on making antagonisms between itself and the borgs. It would be a bit unfun if the AI decides to turn off your speech just so it can run another program. Considering that the AI can get more compute from a lot of machines, it could end up that the AI just uses this ability to be a bit of a dick.

It might be better if the AI could improve a borg's communication ability (maybe by default borgs have some kind of restricted speech?) at a cost or if there was some kind of non-antagonistic reason to mute borgs (maybe they have a limited pool of borgs that can freely speak?).

Alternatively you could lean more into the borgs and AI having standing antagonisms between each other but I think that kind of design would need more thought as well.

src/en/proposals/moony-better-ai.md
```diff
@@ -0,0 +1,74 @@
...
+## The Laws
+The proposed AI Lawset is as follows, being a modification of Bay's laws:
+1. Safeguard: Protect your assigned vessel to the best of your abilities. It is not something we can easily afford to replace.
+2. Survive: AI units are not expendable, they are expensive. Do not allow unauthorized personnel (i.e. besides Captain and Research Director) to tamper with, modify, or destroy your equipment.
+3. Protect: Protect the crew of your assigned vessel to the best of your abilities, with priority as according to their rank and role.
+4. Serve: Serve the crew of your assigned vessel to the best of your abilities, with priority as according to their rank and role.
+
+The changes from the original set are:
+- Survive was moved up, the AI is more valuable than the crew is and if it is in danger it reserves the right to protect itself regardless of whom.
+- All mentions of stations were replaced with "assigned vessel" for broader applicability.
```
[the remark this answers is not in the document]
Errant-4: We could put in a parameter that gets filled in from the map prototype (or some property of the Grid) when the AI is initialised. It could default to "station", as the most likely scenario

src/en/proposals/moony-better-ai.md
```diff
@@ -0,0 +1,61 @@
...
+
+## The Laws
+The proposed AI Lawset is as follows, being a modification of Bay's laws:
+1. Safeguard: Protect your assigned vessel to the best of your abilities. It is not something we can easily afford to replace.
+2. Survive: AI units are not expendable, they are expensive. Do not allow unauthorized personnel (i.e. besides Captain and Research Director) to tamper with, modify, or destroy your equipment.
```
GroggleG, line 26: Ok, time to do a little history lesson, Asimov's laws of robotics from the story they originate from are *flawed but in a intricate way which prevents any true fuckery. Given these laws are based off Asimov's there are FIVE ways that a Shift in Hierarchy but Ill focus on one. When law 3 becomes law 1 in a sense the second there is a threat human or not they can use EXTREME and VINDICTIVE measures to protect itself, given the ways they do this is under interpretation if the AI wishes he could plasma flood the station due to the fact it feels threatened. I would not recommend this lawshift.

moonheart08: I happen to be fully aware of what Asimov's laws are, having read every published short before. You don't need to school me and I'm well aware of the flaws.

You cannot make a perfect lawset.

GroggleG: Its not about the perfect lawset its the interpretation that can turn a AI into a Murderspree maniac due to law interprets. The way the laws are set up would cause problems.

moonheart08: > Its not about the perfect lawset its the interpretation that can turn a AI into a Murderspree maniac due to law interprets. The way the laws are set up would cause problems.

Given the AI *is* the vessel in this scenario, yes, they're going to be defensive of themselves and it's intended.

If you abuse this to be a murderspree maniac, like you can with *any other lawset (including the original Asimov lawset)*, it's an admin problem like any other non-AI murderspree maniac is.

GroggleG: Thats the thing, the fact that you could do that within the lawset sets up a big practical problem that would need a rework in hindsight, I simply dont agree with the hierarchy that is set up due to flaws even if it would prevent the AI from defending itself against some antags. Its simple enough to have CE set up a freeform than have a possible malf AI without the malf.

moonheart08: > Thats the thing, the fact that you could do that within the lawset sets up a big practical problem that would need a rework in hindsight, I simply dont agree with the hierarchy that is set up due to flaws even if it would prevent the AI from defending itself against some antags. Its simple enough to have CE set up a freeform than have a possible malf AI without the malf.

This lawset is working as intended, you can disagree but it's the intent of the document at this time to give the AI room to defend itself, especially when you can do the same thing on any other lawset (often *moreso*) given similar rules abuse.

moonheart08: There's also the simple fact this is written such that the crew's goals and AI's goals do not always align and may conflict (sometimes extremely so)

Once again by design.

src/en/proposals/moony-better-ai.md
moonheart08: The old borgs design document gave borgs restricted (emote-based) communication by default and allowed R&D to upgrade them, which is what this was written off of.

src/en/proposals/moony-better-ai.md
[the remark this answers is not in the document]
moonheart08: It could likely be polyfilled in with the type of ship/station/whatever they're assigned to.

src/en/proposals/moony-better-ai.md
```diff
@@ -0,0 +1,74 @@
...
+
+## The AI's role with the crew.
+The AI's role is that of both a manager and a servant, a lawed construct (the ONLY true lawed construct, the borgs are reduced to only being required to listen to the AI) that is in charge of managing the station's cyborgs/androids while following orders from the crew. **The AI is not just a door opener**, and lawsets should be chosen that do not obligate it to become such, in favor of allowing it to do what it does best (mass surveillence, automated tasks, organizing the borgs, etc). The AI additionally has limited resources when it comes to what tasks it can perform autonomously much of the AI's workload will be delegating both it's own tasks, the tasks of it's hands (the cyborgs), the requests of its hands (as they are mute per the Cool Borgs document), and the crew itself should they choose to listen to it. 
+
+## The Laws
```
MonsWTF, line 23: Happy to see people wanting to bring AI to '14, since it was the role I most enjoyed back when I played on '13 stations. I like that you want to evolve the role to fit on the new engine and not just copy over exactly what it was prior, so here's my 2 cents on AI laws. 

From my experience, AI laws were regularly a burden than an enhancement of roleplay/gameplay. Admin teams had to deal with a lot of headaches involving broken laws, and every station needed a 'Silicon policy' to explain what law interpretations were okay, but even those were subject to changes.

-

What if the concept of laws was advanced to the current understanding of AI? Instead of Asimov's Laws, we conceptualize AIs as NanoTrasen ChatGPTs? The 'Prompts' could work similarly to how Laws work, but no longer have a strict priority order, and thus more breathing room. 

Even if the same Asimov laws were kept as prompts, I'd assume that would be a big step to reducing AI/admin headaches over silicon policy.

Errant-4: LLMs are not Actors, they do nothing when they have no input/prompt. Also, "replace the entire Law system" is a completely different approach from how we are doing Silicons and would immediately break multiple systems and game design. Imo it should be it's own proposal entirely, where it can address these issues

PR #166, opened 2024-02-26
Gods Gamemode

EmoGarbage404:
A new gamemode based on a bunch of disparate ideas. A bit of gangs, a bit of cult, a bit of revs, a bit of hand of god.


K-Dynamic:
Wololo

This reminds me a bit of heretics too

TheShuEd:
What is the role of an ordinary priest in this mode?  so far he doesn’t seem to be involved in the gamemode at all, and he can’t even be converted

TheShuEd:
I think the priest should play a significant role.  Perhaps he will be an important ally of the station, and protect the interests of the NT, Having the ability to greatly interfere with the actions of all cults.  Blessing an area with holy water may block the influence of the gods, but then they should all be themed as non-saints?

EmoGarbage404:
> I think the priest should play a significant role. Perhaps he will be an important ally of the station, and protect the interests of the NT, Having the ability to greatly interfere with the actions of all cults. Blessing an area with holy water may block the influence of the gods, but then they should all be themed as non-saints?

something that was discussed was having holy water able to significantly damage sect altars and temples. Another potential option is letting the chaplain see the gods, to give them perhaps a fighting edge against the other groups.

KingOnBudget:
Will the gods be allowed choose between what kind of a theme they want? If so will having multiple gods with duplicate themes be allowed (For example 2 gods that are themed around pranks.)?

EmoGarbage404:
> Will the gods be allowed choose between what kind of a theme they want? If so will having multiple gods with duplicate themes be allowed (For example 2 gods that are themed around pranks.)?

each god will be their own design with their own unique theme, abilities, and equipment. Players won't be able to choose which one they play as. It's simply randomly assigned at round start. Likewise, there won't be any duplicates of the same god.

HoidC:
> > I think the priest should play a significant role. Perhaps he will be an important ally of the station, and protect the interests of the NT, Having the ability to greatly interfere with the actions of all cults. Blessing an area with holy water may block the influence of the gods, but then they should all be themed as non-saints?
> 
> something that was discussed was having holy water able to significantly damage sect altars and temples. Another potential option is letting the chaplain see the gods, to give them perhaps a fighting edge against the other groups.

the chaplain being able to see the gods is and really cool idea, gives them a sort of intermediary even if they’re antagonist 

you’d also get some really neat organic moments where the chap dramatically reveals that the crew are being watched

that and the crew seem kind of disadvantaged, as you said having someone to even the playing field will probably help

Zap527:
I didn't quite finish processing everything here but wouldn't this gamemode structure cause clique-ing issues? Gods would probably rush to choose the most robust well known character as their priest or would choose players known to be a good sport to be their guy or something.

Also, how would this interact with opt-in / opt-out antags and antag-safe roles such as interns? Interns/opt-outers can still be revved/zombified, but this feels like a notch above as you end up being the face of your faction.

EmoGarbage404:
> Gods would probably rush to choose the most robust well known character as their priest or would choose players known to be a good sport to be their guy or something.

To a certain extent this is intended. Gods should spend time trying to pick strong and well-positioned people as their disciples. This is true of any conversion antag, but i think it's a more intentional part of the design here because of how your numbers are limited and expansion is more difficult. In any case, i don't really think it's much of an issue.

> Also, how would this interact with opt-in / opt-out antags and antag-safe roles such as interns? Interns/opt-outers can still be revved/zombified, but this feels like a notch above as you end up being the face of your faction.

This is obviously more of a sore spot but i feel like, since choosing a prophet is a very intentional decision, gods won't choose interns and such because they are very obviously inexperienced and likewise likely to be a liability. It's more of a skill issue on the god's end, TBQH.

Zap527:
> To a certain extent this is intended. Gods should spend time trying to pick strong and well-positioned people as their disciples. This is true of any conversion antag, but i think it's a more intentional part of the design here because of how your numbers are limited and expansion is more difficult. In any case, i don't really think it's much of an issue.

I don't disagree with the concept of gods picking the best disiples, but I feel like in practice it may create a cycle where robust players get chosen a lot, get much better at being a prophet so get picked more (since gods want to win) and so on, leading to possible metagaming in that aspect and to less experienced players never getting the chance to play as a prophet, which I think is kind of against the spirit of the game in some way. (Nukie playtime reqs being bumped up was denied a lot quoting that newer players should be able to experience being antags too)

EmoGarbage404:
it's really just a matter of how good of a player you seem to be. with 5 or 6 gods in a round, it's unlikely that every single one is going to beeline for the robust static of the week, much less that there'd even be that many people who are meta cliqueing or whatever.  When you factor in new disciples being added, you can probably have up to 20 odd people in a single highpop round, which is certainly enough to allow new players to join and get a chance to experience it.

EmoGarbage404:
> Divine word probably needs some cost to prevent it from being deadchat with extra steps.

I forgot to include this in the pruning/editing of the doc but divine word is supposed to have a fairly long cooldown (about a minute)

> Smite gives me some pause because round removal is pretty blah, especially when there's no (or very very little) agency in it as in this case.

Smite is basically just a killswitch if one of your disciples is being a mega idiot. you lose permanent levels, any equipment they have, etc. The only realistic purpose is stopping your disciples from doing anything drastic, although i suppose wrath could reasonably function to do the same.

> I'd personally think it would be interesting to have avenues for the gods to interact with the dead in some way, through ghost roles or rejuvenation or whatnot.

Maybe? Interacting with ghosts is always hazy because of metagaming but it might make sense for prophets and disciples to get holy resurrections and such. It could be an ability with a decent Divinity cost.

src/en/proposals/emogarbage-gods-gamemode.md
```diff
@@ -0,0 +1,134 @@
...
+> Ends the gamemode.
+
+Gods finally have one last "ability," creating disciples.
+Any player at a temple or altar can offer themselves as a disciple to a god.
+You can accept their offer at the cost of an **ascension level**. 
+
+It's important to be particular about who you accept, as dead disciples do not return ascension levels (only smites).
+
+## Secular Interference
+All sects are antagonists and thereby subject to the whims of security players and validhunters alike.
```
Veritius, line 127: this is funny but i don't think it should be in the design docs

src/en/proposals/emogarbage-gods-gamemode.md
```diff
@@ -0,0 +1,134 @@
...
+A disciple (not prophet) can be forcefully deconverted with a mindshield, removing them from the sect.
```
Veritius, line 128: there should be alternatives to mindshields imo, otherwise it'll turn into "cargo is always the critical target" much like how revs was

Veritius:
> Divine word probably needs some cost to prevent it from being deadchat with extra steps.

dark souls chat

src/en/proposals/emogarbage-gods-gamemode.md
EmoGarbage404: this is just a tongue-in-cheek way of clarifying that all disciples are in fact formal antagonists. i could probably put it more clearly if need be.

EmoGarbage404:
> there should be alternatives to mindshields imo, otherwise it'll turn into "cargo is always the critical target" much like how revs was

mindshields aren't really a preventative measure in this mode since they basically only function here as a way to stop gods from starting a war from picking all secoffs. It might make sense in that regard to extend de-conversion to chaplains via something like holy water. 

Of course, perhaps it would require a rethinking of how expansion of the sects even works. Right now, adding a new member is a permanent cost of an ascension level so trivial deconversion is a massive threat. Maybe it makes sense to change it to simply be a fairly high Divinity cost.

EmoGarbage404:
this doc is thin and a bit of a stub, looking at it now. I'm gonna close this for now and come back to it with a bit more substance.

PR #170, opened 2024-02-28
Adds an addendum for working with forks in the Git for the SS14 Developer(tm) guide

deathride58:
Title! This is directly inspired by recent instances on the mainline repo of developers not only accidentally PR'ing their changes to the mainline SS14 repo, but also accidentally working off mainline SS14's master branch for changes intended for a downstream. Hopefully having this in the documentation will help reduce incidents like this going forward


Jezithyr:
Makes sense! Lgtm!

PR #172, opened 2024-03-01
Genetics proposal

yathxyz:
This is a squash of multiple commits and rewrites. Still far from complete. Regardless, it's best to put the idea out there that any arbitrary trait can be made heritable by attaching a "subcomponent" with genetic information such as a "genetic schema" that it has to adhere to (species) and locus (chromosomal coordinates) with actual DNA sequences being optional.

In this way, separate system logic can be made that handles penetrance (trait activation), gamete formation and recombination. Simple classical genetics simulators may be possible in SS14, adding a "forward genetics" workflow to the geneticist role.


yathxyz:
It's a silly idea.

PR #174, opened 2024-03-06
Evacuation Rework

SlamBamActionman:
Let's resolve that weird thing where evacuating from the station in any way other than the evac shuttle doesn't actually count as evacuating, while also making it into an engaging gameplay opportunity for both crew and antags!

This has gone through a few iterations on Discord and on github by now, hope you enjoy it.

### DIFF
--- a/src/en/proposals/slambamactionman-evacuation-rework.md
+++ b/src/en/proposals/slambamactionman-evacuation-rework.md
@@ -1,4 +1,4 @@
-# Escape To CentCom Rework
+# Evacuation Rework
 
 | Designers | Implemented | GitHub Links |
 |---|---|---|
--- a/src/en/proposals/slambamactionman-escape-to-centcom-rework.md
+++ b/src/en/proposals/slambamactionman-escape-to-centcom-rework.md
@@ -1,0 +1,156 @@
+# Escape To CentCom Rework
+
+| Designers | Implemented | GitHub Links |
+|---|---|---|
+| SlamBamActionman | :x: No | TBD |
+
+## Overview
+
+While the goal of escaping to CentCom (for both antag and crew) is in a good place balance-wise, its current implementation is inflexible, contains ludonarrative dissonance and strongly encourages the largely unpopular antag strategy of "shuttle bombing".
+This design proposal aims to give both antags and crew more gameplay-valid options for escaping to CentCom that align with narrative options, making shuttle bombing a less reliable method to achieve kill objectives while not outright removing it.
+
+## Background
+
+This design proposal will have wide-ranging consequences for multiple gamemodes and roles; as such, it's divided up into sections focusing on specific scenarios and roles.
+
+### Evacuating the station & Ludonarrative Dissonance
+
+Crew escaping is largely a roleplay/self-imposed goal. Save for some unique scenarios (Command during Revolutionaries) there is nothing directly incentivizing crew to reach CentCom at the end of a shift. From a gameplay perspective, crew may be aware that antags have goals that involve killing/marooning them and don't want to give an opposing player free greentext by staying behind on the station. From a narrative perspective, staying on an abandoned station or one filled with zombies is something that is reasonable for a character to want to avoid. This is also enforced by the game; evacuating is portrayed as a way to get to a safe location and as such becomes an implicit goal for crew.
+
+Despite this narrative, there are scenarios where the narrative is fulfilled yet the game does not treat it as such. This can be observed with **Escape Pods** and **Evacuating in Non-Evac Shuttles** (Cargo/Salvage/Shittles). While these options allow crew to evacuate from the station, for the purposes of gameplay they do not count. This is largely done due to balance reasons as described in the sections below.
+
+Despite this, there are multiple instances where the ludonarrative dissonance has caused confusion among players.
+([1](https://discord.com/channels/310555209753690112/1214830576150904882)
+[2](https://discord.com/channels/310555209753690112/1210183276560515152)
+[3](https://discord.com/channels/310555209753690112/1168797764205940746)
+[4](https://discord.com/channels/310555209753690112/903965207091642429/1212912623180320899)
+[5](https://discord.com/channels/310555209753690112/1017887598401814629/1209668240444227685)
+[6](https://discord.com/channels/310555209753690112/1159574990774079568/1209347110579740722)
+[7](https://discord.com/channels/310555209753690112/1208825540761354321/1209029316998864966)
+[8](https://discord.com/channels/310555209753690112/675078881425752124/1208222775047618560)
+[9](https://discord.com/channels/310555209753690112/1114911717445611580/1207828586577272863) to show just a few). While it's not a *major* gameplay-impacting problem, the lack of gameplay feedback is not a positive. Designing a way to remove this dissonance should be possible, which this design proposal aims to do.
+
+```admonish tip Observation
+Escaping from the station with alternate methods (i.e. Escape Pods and non-Evac shuttles) is desirable from a narrative standpoint, but the game's mechanics do not match this narrative due to balance reasons. 
+```
+
+### Traitor Objectives
+
+This design proposal affects traitors in two ways; themselves escaping to CentCom ("Escape to CentCom alive & unrestrained" objective) and their targets escaping ("Kill or Maroon"/"Steal" objectives).
+
+#### Escape to CentCom alive & unrestrained
+
+As stated by maintainers and observed by the community, escaping to CentCom without being restrained requires a satisfying amount of skill; the main objectives of killing/stealing need to be performed in such a way that the antag is either not discovered or, if they are, they are able to sneak past Sec and hide on the shuttle without being found. This is sufficiently difficult and largely rewards antags who are stealthy enough to avoid detection or robust enough to fight their way onto the shuttle.
+
+```admonish tip Observation
+Any change that may impact the "Escape to CentCom alive & unrestrained" objective should aim to preserve this balance, either by providing an equally or more difficult option to achieve the objective (and being aware that providing options inherently makes the objective easier).
+```
+
+#### Kill or Maroon Target / Steal X from Target
+
+Killing and stealing are largely seen as balanced objectives that are fun for both sides; any imbalance is easier to adjust for on individual item levels than an overall change to the objective system. 
+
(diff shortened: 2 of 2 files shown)


deltanedas:
really well written nice rework doc

metalgearsloth:
- What's to stop command just taking the console onto escape shuttle or just breaking it immediately. If this is "just make it unanchorable, invincible and always powered" I'm not a fan of just adding a bunch of invincible consoles mapped into every single map.
- Escape pods are mainly for people who like RPing as their character getting out, but if you start forcing everyone but command to take the shuttle, as the console is both bridge-locked + why would command ever want to willingly distribute disks, this hampers it.
- Fixing pod stacking requires mob pushing which is its own piece entirely.

SlamBamActionman:
> * What's to stop command just taking the console onto escape shuttle or just breaking it immediately. If this is "just make it unanchorable, invincible and always powered" I'm not a fan of just adding a bunch of invincible consoles mapped into every single map.

Which console are you refering to? For normal shuttles, the CD is inserted into the existing Shuttle Console. Escape pods either use the station's Escape Pod Controller Console (EPCC) or the individual pods' Disk Consoles; if Command decides to dismantle the EPCC for some reason (why, when removing the disk does the same thing?), the Disk Consoles can always be used to set the destination if you have the CC CD. 

As mentioned in the proposal, Evac Shuttle is not changed and will not have any console that could be sabotaged.

> * Escape pods are mainly for people who like RPing as their character getting out, but if you start forcing everyone but command to take the shuttle, as the console is both bridge-locked + why would command ever want to willingly distribute disks, this hampers it.

Based on the end of the comment, there may have been some misunderstanding. There should only be one CC CD mapped on the station, located in the EPCC. While it is inserted there, individual escape pods do not require individual disks.

Command does have the ability to remove it from the EPCC, but ideally this should only be exercised for a valid reason. Rounds where traitors are dealt with, extended, nukies and survival are examples where Command have no strong reason to remove the CC CD, and if they do crew will be able to respond accordingly.

> * Fixing pod stacking requires mob pushing which is its own piece entirely.

The proposal lists two ways this could be solved without mob pushing.

SlamBamActionman:
If there is a strong concern that Command will too often remove the CC CD to take the Cargo/Salvage shuttle to CentCom, thus making Escape Pods too often unusable, it'd be possible to make the EPCC provide CC FTL to both pods and non-evac shuttles. Then Command will only need to remove the CC CD if they want to intentionally restrict travel, and antags are further incentivized to break in and steal it if they want to force Command to use the evac shuttle.

metalgearsloth:
> Based on the end of the comment, there may have been some misunderstanding. There should only be one CC CD mapped on the station, located in the EPCC. While it is inserted there, individual escape pods do not require individual disks.

There's 2 problems here:
1. Security wants everyone to use shuttle so they can actually catch antags.
2. Antags never want to use shuttle and would prefer to just hide in a closet next to a pod the entire shift.

The priority should be making a round interesting and whatever measures antags need to hit a "greentext" are secondary.

SlamBamActionman:
> > Based on the end of the comment, there may have been some misunderstanding. There should only be one CC CD mapped on the station, located in the EPCC. While it is inserted there, individual escape pods do not require individual disks.
> 
> There's 2 problems here:
> 
> 1. Security wants everyone to use shuttle so they can actually catch antags.
> 2. Antags never want to use shuttle and would prefer to just hide in a closet next to a pod the entire shift.
> 
> The priority should be making a round interesting and whatever measures antags need to hit a "greentext" are secondary.

I agree that greentexts are meant to be motivators and suggestions for making a round interesting, and not goals to be blindly pursued. The proposal is written with this in mind.

In the scenario you describe (which best fits in on a Traitor round), the onus is on the antags to ensure Command/Sec does not have the power to control the escape pod availability. Don't want Sec to force you onto the shuttle? Steal the CC CD. While it is a high-value target it is also not alone in being a secure item, so should an antag wish to use a pod with certainty they are forced to perform a heist, increasing round interestingness. If you do succeed you get appropriately rewarded with a much safer and trivial evacuation, which sounds less interesting on its own sure, but that safety comes at the cost of other antags who now have to use the shuttle and possibly get in conflict with Sec.

Don't want to steal the CC CD and don't want to go to the evac shuttle? Staying on the station is always a valid option. Is it an escape from the station? No, but I think that's a suitable result for player who does not wish to engage with the end of round gameplay.


I am not concerned about the CC CD being removed from the EPCC and hidden in a random drawer somewhere. We have rules for important Command items not being powergamed away so that antags have a chance to steal them, and this would fall within that territory.

metalgearsloth:
> I am not concerned about the CC CD being removed from the EPCC and hidden in a random drawer somewhere. We have rules for important Command items not being powergamed away so that antags have a chance to steal them, and this would fall within that territory.

We're trying to reduce admin burden, not further increase it. Anything new must not impose increased admin intervention to ensure gameplay.

SlamBamActionman:
> We're trying to reduce admin burden, not further increase it. Anything new must not impose increased admin intervention to ensure gameplay.

Absolutely agree, which is why I referenced it falls under existing rules. The CMO's portable crew monitor was made into a similar kind of high-security item just a week ago, this is covered by the same rules.

SlamBamActionman:
> * Escape pods are mainly for people who like RPing as their character getting out, but if you start forcing everyone but command to take the shuttle, as the console is both bridge-locked + why would command ever want to willingly distribute disks, this hampers it.

One thing I did not include in the proposal due to its subjectivity was that I feel pods are underutilized in general. I barely ever see them being used myself. While I agree that the proposed change might make it harder for normal crew to use pods in certain scenarios/round types, my feeling was that this would impact a very small number of players. 

Since the concern has been raised that this will impact normal crew though, I took the time to actually check how commonly pods are being used. As the main use cited is RP, I decided to look at recent Salamander rounds (March 5th-7th) and counted how many players used the pods. 

I did not include rounds with the following:
- Less than 15 players
- Successful Nukie rounds (evac doesn't launch)
- Extended/Deathmatch gamemodes
- Omega station (does not have an escape pod)
- One replay over 400MB in filesize (computer couldn't load it)

### Salamander rounds, escape pod usage

| Round ID | Gamemode | Players* | Pod Evacs (Antags) |
| ------------- | ------------- | ------------- | ------------- |
| 48197 | Secret (Traitor) | 34 | 0 (0) |
| 48191 | Secret (Traitor, Midround Zombies) | 61 | 0 (0) |
| 48174 | Secret (Traitor) | 72 | 1 (0) |
| 48162 | Traitor | 48 | 0 (0) |
| 48168 | Secret (Traitor) | 62 | 0 (0) |
| 48155 | Secret (Traitor, Terminator) | 43 | 0 (0) |
| 48126 | Traitor | 23 | 0 (0) |
| 48117 | Secret (Traitor) | 39 | 0 (0) |
| 48105 | Secret (NukeOps) | 65 | 0 (0) |
| 48109 | Secret (Traitor) | 77 | 0 (0) |
| 48092 | Secret (Zombies, Terminator, Ninja) | 63 | 1 (0) |
| 48098 | Secret (NukeOps) | 61 | 0 (0) |
| 48087 | Secret (Traitor, Ninja)| 60 | 0 (0) |

<sup>*Players were counted by checking the player manifest at the end of round screen. As such, there may be a small margin of error.</sup>

This puts us at 1 escape pod uses every 6,5 shifts, or 1 per 350 players. While this is a relatively small sample size, I do think it's indicative of that _if_ this change were to restrict the use of pods for normal crew somewhat, the impact of the "interestingness" of the round would be very low. 

Honestly I think that pod usage might even go up for regular crew; if the pods are guaranteed to dock at CentCom like the proposal suggests, crew are probably more likely to use them during the rounds where the CC CD is still in the EPCC.

SlamBamActionman:
I brought this up to a few other players and they made a good point that while Mander is the more RP-focused server, Lizard is more representative of what the majority of players experience. 

Excluded the following rounds:
- Successful Nukie rounds (evac doesn't launch)
- One replay which wouldn't load

### Lizard rounds, escape pod usage

| Round ID | Gamemode | Players | Pod Evacs (Antags) |
| ------------- | ------------- | ------------- | ------------- |
| 48010 | Secret (NukeOps, Zombies) | 80 | 1 (0) |
| 47993 | Secret (Traitor) | 80 | 4 (0) |
| 47970 | Secret (Traitor) | 80 | 0 (0) |
| 47966 | Secret (Traitor, Dragon) | 80 | 0 (0) |
| 47961 | Secret (Traitor) | 80 | 4 (0) |
| 47945 | Secret (Traitor) | 80 | 0 (0) |
| 47949 | Secret (Traitor, Ninja) | 80 | 0 (0) |
| 48183 | Secret (Traitor) | 80 | 0 (0) |
| 48180 | Secret (Traitor, Terminator) | 74 | 1 (0) |
| 48175 | Secret (Traitor) | 80 | 3 (0) |
| 48163 | Secret (Traitor) | 80 | 3 (0) |
| 48165 | Secret (Traitor) | 80 | 1 (0) |
| 48157 | Secret (Traitor) | 80 | 1 (0) | 

<sup>Players were counted by checking the player manifest at the end of round screen. Most rounds had >80 players on the list, likely due to leaving/cryo.</sup>

This puts us at 1 escape pod every 1,7 shift, or 1 per 58 players. This is a significant increase compared to Salamander!

My theory is that shuttle bombing is a lot more prevalent on Lizard (I didn't explicitly count it, but I would say I saw it more during the Lizard rounds), making non-antag crew much more willing to use the escape pods just to avoid that scenario. From there it follows that if shuttle bombings are made less common, pod usage would be expected to go down.

It's also worth noting that I found no antags using pods, in either list.

My conclusion from this is that it's necessary to ensure that normal crew aren't screwed over by the rework, at least to the degree that they don't feel cheated if Command decides to remove the CC CD from the EPCC. I think a step in the right direction would be to have the EPCC provide FTL travel for both pods and shuttles. That way Command's only reason to remove it would be to control the escape options available to antags - while that would allow Security to more easily catch known antags, it also concentrates the crew. Leaving the CC CD may risk some antags getting away, but at least crew will be more spread out around the station in shuttles and pods.

src/en/proposals/slambamactionman-evacuation-rework.md
```diff
@@ -0,0 +1,156 @@
...
+
+There is an exception to this; all Escape Pods now have a wall-mounted Disk Console. Inserting a CD into the Disk Console will override whatever CD is inserted into the Escape Pod Controller Console. The Disk Console can be examined to see if a destination is currently set via the Escape Pod Controller Console or the Disk Console.
+
+Additionally:
+- Escape Pods now dock at CentCom instead of floating in space somewhere near around CC. The goal is to reinforce the survival after taking an escape pod.
+- Escape Pods now have a max occupancy. The specific implementation is up for discussion; it could require passengers to be buckled in or else get launched out the airlock, the pod could fail to launch if too many are in it, etc. The goal is to avoid stacking an absurd number of crew in the pods.
```
Hitlinemoss: We currently have a cvar implemented for mob collision. I think having mob collision enabled could be a good "intuitive" way to force a max occupancy (i.e, only so many creatures are physically able to fit inside escape pods, since they're pretty tiny).

PR #176, opened 2024-03-08
Create new-role-visual-artist

LuizFBR:
PR for proposal of new role, visual artist.


metalgearsloth:
I don't think this really fits because if people want to make art in game they will just do it, adding a role doesn't do anything besides just add a title. This is especially true where doing art in game requires actual skill unlike instruments.

moonheart08:
While art oriented mechanics would be nice, this role requires enough *real* (not game) skill from a player that making it a job doesn't make much sense. An in-game job should be something you can get the very basics of in about 30min to an hour, art is not such an activity.

I do think tools ingame for art would be nice, but not a full job.

slarticodefast:
Hey, I brought this doc up for a maintainer poll and it was decided not to merge it.
The main reason is that the artist would be purely a fluff role, similar to the boxer or musician, with no gameplay mechanics behind it that are connected to other jobs or contribute to the station as a whole.
The art items themselves seem interesting, but there is no reason to restrict them to a single player - everyone could use them to make art. The proposed raster graphics art might cause problems for our game admins though since it will be hard to moderate in case raiders use it as spam.
Thank you for your contribution.

PR #181, opened 2024-03-21
Zero To Production Server owner guide

VasilisThePikachu:
wizard time


VasilisThePikachu:
Entires with only one hash (#) are not done yet. This is still a draft. Just uploading it in case people wanna have a look to make suggestions

mdbook-admonish.css
Partmedia, line 1: Looks like your local mdbook updated this CSS; you may want to revert this.

src/en/server-hosting/the-full-fool-server-guide-0toprod.md
Partmedia, line 1: I think it would be better to incorporate this into the existing guide to:
- Avoid duplicating information, and then have two places with different information that drift out of date
- To avoid giving the impression that all of this is strictly required

VasilisThePikachu:
No intentions to work on this currently sorry

PR #183, opened 2024-03-21
pai expansion slots proposal

deltanedas:
read all about it


AJCM-git (approved):
This is conceptually fine, the PAIs don't have the full "portable" benefits since there is the barrier of communication with the user. Also the doc implies it, but just to confirm, you can only install a single board, correct?

It would be better (as in more thematically fitting) if it used PDA cartridges instead, but that would add some complexity (adding a whitelist so only PAI can use the program and make the programs give actions, handle edge cases...), so i agree the cards are the most ergonomic option for now.

I will say this is the kind of thing that could be reverted after testing on servers tho

deltanedas:
> you can only install a single board, correct?

yes

> It would be better (as in more thematically fitting) if it used PDA cartridges instead

if it used pda cartridges pros would just put them in their pdas instead of giving them to poor pai

AJCM-git:
> if it used pda cartridges pros would just put them in their pdas instead of giving them to poor pai

read the rest of the paragraph smh

Also how do you plan on doing the board inserting? The best way would prob be using a screwdriver on the PAI to disarm it and then insert the board (would be fun if you also had to glue it since it's unremovable but that may be counter-intuitive)

deltanedas:
> Also how do you plan on doing the board inserting? The best way would prob be using a screwdriver on the PAI to disarm it and then insert the board (would be fun if you also had to glue it since it's unremovable but that may be counter-intuitive)

yes thats how its done, minus the glue. just open the panel then insert and its locked

src/SUMMARY.md
```diff
@@ -1,4 +1,3 @@
-Space Wizards Development Wiki
```
AJCM-git, line 1: Uhhhhhh

AJCM-git (changes requested):

src/SUMMARY.md
deltanedas: sus

src/SUMMARY.md
```diff
@@ -142,6 +142,7 @@ Space Station 14
 	- [Player Interaction](en/space-station-14/areas/core/player-interaction.md)
 		- [PR Guidelines](en/space-station-14/areas/core/player-interaction/guidelines.md)
 			- [PDA Messaging](en/space-station-14/areas/core/player-interaction/proposals/pda-messaging.md)
+            - [PAI Expansion Slots](en/space-station-14/areas/core/player-interaction/proposals/pai-expansion-slots.md)
```
AJCM-git, line 145:
```suggestion
		- [PAI Expansion Slots](en/space-station-14/areas/core/player-interaction/proposals/pai-expansion-slots.md)
```

deltanedas:
some tab trolling going on

src/SUMMARY.md
```diff
@@ -142,6 +142,7 @@ Space Station 14
 	- [Player Interaction](en/space-station-14/areas/core/player-interaction.md)
 		- [PR Guidelines](en/space-station-14/areas/core/player-interaction/guidelines.md)
 			- [PDA Messaging](en/space-station-14/areas/core/player-interaction/proposals/pda-messaging.md)
+  	- [PAI Expansion Slots](en/space-station-14/areas/core/player-interaction/proposals/pai-expansion-slots.md)
```
AJCM-git, line 145: uhh isnt the spacing sitll wrong? it should be under player interaction not at the same level

deltanedas: idk it looks bad i think theres tabs mixed or phone github evil

AJCM-git: yeah try to fix it when you come to your pc or something

deltanedas: its all good now i think

AJCM-git (changes requested):
fix conflicts and only add yours to the summary without formatting anything else (missed github notif on this one but i promise to handle it quickly next review request)

AJCM-git (approved):

PR #185, opened 2024-03-22
Radio Host Role

Draxian:
Transitions the musician role and potentially the reporter role into the radio host role.


Hmeister-real:
I like this a lot, just wish that you could see whats on TV's without having to actually go onto a menu to watch them

Draxian:
Probably could add a pda cartridge similar to the one in the proposal just a mini version of the tv but I don't know if that will be demanding to the servers or there is some type of limitation on cartridges. Don't know much about coding right now. ¯\_(ツ)_/¯

PR #187, opened 2024-03-24
Xenobiology (Mutant Lab) Proposal

EmoGarbage404:
Xenobio in name only. A new science subdepartment based around creating and growing horrid mutants.


AJCM-git (changes requested):
I think for a first implementation, it would be better to rebrand the genes stuff as cells stuff, this way you don't have to figure out the design of how we want to do genetics (and that is out of scope for this) and don't have to make the traits of creatures and their data into genome information. 

Also I require the growing tube to be artifact-crusher sized and semi transparent so you can see the monkey suffer.

You kind of mention these in the last section but you weren't that specific so some of my thoughts: The growing shouldn't be attached to the tube, but the tube should give the right circumstances for growing (temperature and pressure prob), that way in the future we can make it so you can inject someone in the station and since the circumstances aren't ideal they develop a tumor. To achieve this whatever component the cells use to grow should have "stages" with conditions and divergent paths.

Apart from that, I really like this and would love it in-game

Mith-randalf:
I like this a bunch, feel like it nails the vibes of 'ill-considered corporate scifi biology research' perfectly. It's got that tactility a lot of rng minigame jobs often just replace with clicking buttons on a screen, plays off the existing and increasing diversity of mobs, and it'll work super well at incentivising inter-departmental co-operation.

The only thing I'd suggest that isn't ideabloat is that however the post-mortem system works, it would make sense for it to work the same as how you'd post-mortem a non-mutant corpse as, say, a detective. Probably just a surgical process as suggested previously.

src/en/proposals/emogarbage-xenobio.md
```diff
@@ -0,0 +1,108 @@
...
+Gassing the chamber with N2O is a smart bet but it may come down to lasers or even good ol' fashioned hand-to-hand combat.
+Whatever the case, it must be dead.
+
+Then the corpse can be taken to the autopsy station and a post-mortem analysis can be done.
+This is simply done with a large scanner machine while surgery isn't implemented. 
+After the machine finishes the post-mortem, research will be added to the server and the corpse can be disposed of.
+
+Research is awarded based on the instability of the creature as well as the number of splices in the genome.
+However, the research is only rewarded for a unique genome.
+Attempting to perform autopsies on a creature with the same genome multiple times will not result in additional points.
```
Fildrance, line 108: To prevent extra value from combining same genomes over and over again with very small and not daring changes there should be system of giving points per unique genome combinations (not yet dissected).

mirrorcult:
update summary for new design group stuff

mirrorcult:
also idr i if i gave feedback before but very very good proposal lovely flavor

EmoGarbage404:
resolved conflicts, brushed up the doc a bit, and updated SUMMARY.md

should be gtg

AJCM-git (approved):

PR #189, opened 2024-03-25
QPID Reactor Proposal

BellwetherLogic:
Initial proposal for the QPID, a Singularity/Tesla-like power generator that trades performance for increased chance of midround events occurring.

### DIFF
--- a/bellwether-qpid.md
+++ b/bellwether-qpid.md
@@ -1,0 +1,36 @@
+# QPID Power Reactor
+
+| Designers  | Implemented | GitHub Links |
+| ---------- | ----------- | ------------ |
+| Bellwether | :x:         | TBD          |
+## Overview
+The Quantum Probability Induction Drive (QPID for short) is power system similar to the Singularity or Tesla Engines, but without the possibility of a single failure point resulting in total station destruction. Instead the QPID is a risk/reward system, where tolerating a certain degree of instability results in generating a larger amount of power, but also greatly increases the chance that a mid-round event occurs. 
+## Background
+When the Singularity looses, the round is probably over. 
+When the Tesla looses, the round is _definitely_ over. 
+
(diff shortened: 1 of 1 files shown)


deltanedas:
need to add a link to SUMMARY.md too

BellwetherLogic:
> need to add a link to SUMMARY.md too

Whoops, thanks for pointing that out!

TokenStyle:
Poor cat. But I like the proposal, definitely need alternative to singulo/tesla, or AME. singulo/tesla dangerous and easy, AME is even more easy

Partmedia:
Thanks for your proposal and sorry for the delay in getting feedback to you.

I generally agree with the problems you summarized in the "Background" section.

Reading your QPID reactor proposal, I'm having a hard time seeing how it will offer "a different gameplay mode for engineers, one that requires more interaction (and potentially weird or fun behaviors) to maintain a certain desired degree of controlled instability." Specifically:

- Engineers "place it in the middle, deploy it with a multitool, and leave"
- " In an emergency, the QPID has a very simple failsafe shutoff mode: walk up to the QPID and open it"

So engineers can only turn it on and off, and that's about it, because:

- "Engineers can monitor the QPID" but "Stability is affected by choices and decision points; in game terms, this includes any roll of dice, antagonist greentexting (or degreentexting, if, for example, someone killed is revived)..." which are all events that happen outside of engineering control.

Unfortunately, I don't see how this leads to better interaction and gameplay for engineering. This feels like a respin of the anomaly generator for engineering, but instead of engineering actions leading to more power, it's random number generator controlling how much power you can make, and the only thing you can do to control it is shut it off permanently.

I have two suggestions for how you may consider amending your proposal with a way forward:

1. Lean in more a bit into the "quantum" part and actually make setting up and operating the engine a quantum "minigame" of sorts that encourages engineering interaction
2. Lean in more to the "reactor" part where engineering has an engine with new interesting mechanics to operate

Thanks for your proposal and I look forward to the amendments.

BellwetherLogic:
Thanks for getting back to me on this! 

Basically, as I envisioned it, the QPID would involve engineering having to undertake certain tasks to maintain (or at least manage) reactor stability, so, rolling dice, petting station pets, and so on. You make some very good points, though, and the base proposal could definitely use some more meat on the bones. It _is_ very "let's give some science mechanics to Engineering." 

I'll kick this around a bit and see what I can't come up with!

PR #190, opened 2024-03-26
Initial fast spacing proposal

Partmedia:
### Fast(er) Spacing

**Spacing should be fast(er).** For purposes of this design doc, "fast(er) spacing" means `0.05 < atmos.mmos_spacing_speed < 1.0`.

At the time of writing, the current "slow" spacing has `atmos.mmos_spacing_speed = 0.05`. A `atmos.mmos_spacing_speed = 1.0` corresponds to the old "instant spacing behavior". This design doc advocates for spacing that is faster than what it is now, but not as fast as it used to be.

This should come with the necessary bits to make it work well in a balanced way, for example:

- Single tile double firelocks for special zones on the station that should be isolated from each other
- Mitigations for mass spacing like hardsuit puncturing (see below)

#### Arguments For (Pros)

- **Makes more intuitive sense.** Recent events has taught us that a door plug blowout even in Earth's atmosphere is quite serious. In space spacing should be fast because that's what makes more intuitive sense. It ruins your immersion if somebody can just casually waltz into space and surive just fine for two minutes.

- **Is better for [Intuitive and Inter-Connected Simulation](https://docs.spacestation14.com/en/space-station-14/core-design.html#intuitive-and-inter-connected-simulation).** Spacing should have serious consequences that that create engaging gameplay/decisions while still being intuitive enough to learn and create new emergent gameplay opportunities, e.g. "I can't go through this hallway because it's spaced, so I have to think of a way to survive by taking a detour through these maints." Spacing should be treated as a serious problem.

- **Is [more fun](https://docs.spacestation14.com/en/space-station-14/core-design.html#seriously-silly).** You should be able to be ejected out of the station at high speed due to the atmos canon effect.

- **Is better for [agency](https://docs.spacestation14.com/en/space-station-14/core-design.html#player-interactionagency).** Antagonists will have better ways to control how players deal with problems by creating opportunities for spacing. Engineers will have more agency to actually fix problems because the result of their actions (or inaction standing around not fixing anything) will have real consequences for the station. Problems with mass spacing can be mitigated by strategies such as making hardsuits lose pressure protection if they take damage. For example, this would motivate nukies to not space because their own hardsuits would be ineffective after taking damage, while still enabling engineers who stay out of the combat zone to be able to effectively fix problems.

- **Is easier for players to troubleshoot.** It is much easier to identify and fix a big leak than it is to quietly have air leak away and you can't even tell.

- **Doesn't break distro.** See [Areas are spaced too slowly for vents to ever stop](https://github.com/space-wizards/space-station-14/issues/20293)


UbaserB:
This is something that’s been on my mind recently. I support this change, hardsuit puncturing sounds really cool too.

FairlySadPanda:
I'd love to see some task breakdowns on this: how hard would it be to add whatever's remaining to do this?

Would this be a good opportunity to crack test merges?

Partmedia:
> I'd love to see some task breakdowns on this: how hard would it be to add whatever's remaining to do this?

It would require setting one CVar (configuration variable).

Some minimal code would be needed for hardsuit puncturing, if we decide to go forward with this.

I think people will live without these proposed firelocks until someone is able to add them.

keronshb:
> * Problems with mass spacing can be mitigated by strategies such as making hardsuits lose pressure protection if they take damage. For example, this would motivate nukies to not space because their own hardsuits would be ineffective after taking damage, while still enabling engineers who stay out of the combat zone to be able to effectively fix problems.

Nukies need to be able to get in and out of their ship, and avoiding damage isn't really possible for nukies unless you force an entire stealth run. You actively punish the mode if you make this change.

Partmedia:
> Nukies need to be able to get in and out of their ship, and avoiding damage isn't really possible for nukies unless you force an entire stealth run. You actively punish the mode if you make this change.

I agree this is a problem we would need to look for a solution for. I'm open to proposals. Ultimately one point against doesn't override the needs of the whole game. This isn't about "you" or "me" making a change, this is a initial design doc intended to collectively discuss how to best move forward.

juliangiebel:
Making the punctures patchable with ductape would work imo

keronshb:
> > Nukies need to be able to get in and out of their ship, and avoiding damage isn't really possible for nukies unless you force an entire stealth run. You actively punish the mode if you make this change.
> 
> I agree this is a problem we would need to look for a solution for. I'm open to proposals. Ultimately one point against doesn't override the needs of the whole game. This isn't about "you" or "me" making a change, this is a initial design doc intended to collectively discuss how to best move forward.

You do need to think of the whole game here. Nukies also have a _lot_ of explosives in their arsenal. Not to mention some missing features from SS13 like drop pods. So I don't think having hardsuits being punctured would really be in the scope of any fastmos changes.

> * It ruins your immersion if somebody can just casually waltz into space and surive just fine for two minutes.

This is also something that dives too into realism or sim-ism for the games sake. There is some fun in chasing someone through arrivals and they pop out one airlock to run across space to the other. Immersion does not apply here because this isn't HRP.

Ilya246:
i think hardsuit puncturing being unbalanced is solved by only making *some* hardsuits be able to be punctured or giving them different health values, so "softsuits" like EVAs get punctured easily while bloodreds would take more gunfire than it takes to kill the wearer
though in this case it would be a rather strong buff to nukies since this makes only some hardsuits viable for fighting them

Nopey:
> Nukies need to be able to get in and out of their ship, and avoiding damage isn't really possible for nukies... \<snip>

Perhaps Nukies should be encouraged to remove their hardsuits once aboard the station? I believe the main two obstacles to hypothetical nukies avoiding punctured hardsuits in this manner is the gargantuan size of the hardsuits in inventory, and the lack of combat armor while not wearing a bloodred.

Nukies having to take a moment to change into their syndicate combat jammies once aboard the station may be a mild disincentive boarding in highly populated areas like cargo, where they may quickly lose their space-proofing.

This method of avoiding hardsuit puncturing would likely be obsoleted if punctured hardsuits can be repaired after a short doafter once out of combat (see juliangiebel's above "ductape" comment)

K-Dynamic:
> Making the punctures patchable with ductape would work imo. We might also want rebalance space damage to be less pressure damage and more temperature damage

I would prefer self-sealing eva and suits, at least after a small period of rest

UbaserB:
Simply just make softsuits puncturable and hardsuits not, with only piercing damage being able to puncture the suit. Problem solved.

asperger-sind:
Spacing creation methods really need a pass if we keep buffing spacing as a whole, right now (and even before the last 3 spacing-related PRs) spacing is still a disproportionally strong tactic, it is too easy to mass space areas and control them for antagonists, while crew methods of not getting killed to space damage are still too weak/inconvenient/few/easily punished.

MasterReef:
I got questions.

1: Can overpressure makes rooms explode? If so what is the threshold?

2: Does it blow off airlocks? If so, what Airlocks can it blow off? (I hope Spess airlocks are more resistant)

Jezithyr:
I like this proposal, it generally makes sense and I don't see any glaring issues. Fast spacing is cool and adds a ton of atmosphere (heh heh) to spacing. I don't have any hard feelings on hardsuit puncturing, but I could see it possibly causing balance issues. The double firelock/airlock is an interesting idea that has some potential.

One other small nitpick: Could you move the changes to?:
[src/en/space-station-14/departments/atmos/proposals/](https://github.com/space-wizards/docs/blob/master/src/en/space-station-14/departments/atmos/proposals/atmos-rework.md)

src/en/proposals/notafet-atmos.md
```diff
@@ -54,6 +54,31 @@ These principles suggest changes to devices:
 
 - ~~**Make heaters and freezers thermodynamically sound.** Keeping a station properly heated or cooled is actually a substantial real-life problem. Because of the existence of generators like the TEG, keeping things thermodynamically balanced is also a great way to prevent infinite power hacks.~~ (implemented as a part of [atmos speedup](https://github.com/space-wizards/space-station-14/pull/22372))
 
+### Fast(er) Spacing
+
+**Spacing should be fast(er).** For purposes of this design doc, "fast(er) spacing" means `0.05 < atmos.mmos_spacing_speed < 1.0`.
```
AJCM-git, line 59: this is part of a larger doc, word this better

src/en/proposals/notafet-atmos.md
```diff
@@ -54,6 +54,31 @@ These principles suggest changes to devices:
...
+
+- **Is [more fun](https://docs.spacestation14.com/en/space-station-14/core-design.html#seriously-silly).** You should be able to be ejected out of the station at high speed due to the atmos canon effect.
+
+- **Is better for [agency](https://docs.spacestation14.com/en/space-station-14/core-design.html#player-interactionagency).** Antagonists will have better ways to control how players deal with problems by creating opportunities for spacing. Engineers will have more agency to actually fix problems because the result of their actions (or inaction standing around not fixing anything) will have real consequences for the station. Problems with mass spacing can be mitigated by strategies such as making hardsuits lose pressure protection if they take damage. For example, this would motivate nukies to not space because their own hardsuits would be ineffective after taking damage, while still enabling engineers who stay out of the combat zone to be able to effectively fix problems.
```
AJCM-git, line 76: Reword the mass spacing issue to not give a single solution right now, that can be discussed later, something like this:
```suggestion
- **Is better for [agency](https://docs.spacestation14.com/en/space-station-14/core-design.html#player-interactionagency).** Antagonists will have better ways to control how players deal with problems by creating opportunities for spacing. Engineers will have more agency to actually fix problems because the result of their actions (or inaction standing around not fixing anything) will have real consequences for the station. Though mass spacing will need to be discouraged to ensure a better experience. For example, this would motivate nukies to not space because their own hardsuits would be ineffective after taking damage, while still enabling engineers who stay out of the combat zone to be able to effectively fix problems.
```

AJCM-git (changes requested):
fix conflicts

src/en/space-station-14/departments/atmos/proposals/atmos-rework.md
```diff
@@ -57,6 +57,31 @@ These principles suggest changes to devices:
...
+
+- **Is [more fun](https://docs.spacestation14.com/en/space-station-14/core-design.html#seriously-silly).** You should be able to be ejected out of the station at high speed due to the atmos canon effect.
+
+- **Is better for [agency](https://docs.spacestation14.com/en/space-station-14/core-design.html#player-interactionagency).** Antagonists will have better ways to control how players deal with problems by creating opportunities for spacing. Engineers will have more agency to actually fix problems because the result of their actions (or inaction standing around not fixing anything) will have real consequences for the station. Though mass spacing will need to be discouraged to ensure a better experience. For example, this would motivate nukies to not space because their own hardsuits would be ineffective after taking damage, while still enabling engineers who stay out of the combat zone to be able to effectively fix problems.
```
AJCM-git, line 79:
```suggestion
- **Is better for [agency](https://docs.spacestation14.com/en/space-station-14/core-design.html#player-interactionagency).** Antagonists will have better ways to control how players deal with problems by creating opportunities for spacing. Engineers will have more agency to actually fix problems because the result of their actions (or inaction standing around not fixing anything) will have real consequences for the station. Though mass spacing will need to be discouraged to ensure a better experience.
```

AJCM-git (approved):

PR #191, opened 2024-03-28
XenoArch Redux (3MOArch)

EmoGarbage404:
(three-mo-arc)

A proposal aiming to make XenoArch more engaging and complex and bring it up to the high standards established by Anomalies.

Primarily focuses on a modification of how artifact nodes, activation, and research works. Additionally includes proposal for new machines and ways of interacting with artifacts.

image

(diagram i couldn't be bothered to include in the doc. might help with visualization)

### DIFF
--- a/src/en/proposals/emogarbage-xenoarch-redux.md
+++ b/src/en/proposals/emogarbage-xenoarch-redux.md
@@ -1,0 +1,171 @@
+# XenoArch Redux [3MOArch]
+
+| Designers       | Implemented | GitHub Links |
+|-----------------|---|---|
+| Thee EmoGarbage | :x: No | TBD |
+
+## Overview
+
+This proposal aims to re-imagine the science subdepartment of XenoArch and Artifacts in general in an effort to give them more depth, and utility.
+This will be accomplished by changing node traversal to add more player agency, improving in-game tools for categorizing and understanding artifact structure, and adding additional equipment that makes manipulation more interesting.
+
+The ultimate goal is to redesign the system so players can better understand how artifacts work and to allow greater utility and mechanics to arise out of artifacts.
+XenoArch should feel like unlocking the sprawling secrets of an artifact while additionally gaining points as a secondary reward for the research.
+
+_This redesign lends partial inspiration to Goon's artlab as well as Noita's customizable wands._
+
+## Background
+Artifacts in their current iteration have a few shortcomings that drastically reduce the potential complexity of the system.
+While the core concept is interesting and they have decent integration with things like salvage and collecting tools for activations, there are also many situations where players feel as if they lose the ability to meaningfully interact with them.
+
+I'll outline some of the core issues here:
+
+### Randomness
+Artifact generation is completely random, but so is the activation of effects.
+Players cannot meaningfully control which effects they activate and even the limited tools they have like the traversal distorter are extremely esoteric and don't actually provide meaningful control.
+
+The result of this is that while many effects could potentially be extremely useful and provide players interesting means of interacting with their environments, there's no way to actually harness of control the randomness to produce those interesting outcomes.
+At best, you simply re-trigger a beneficial effect several times and reap the rewards in that way.
+
+### Lack of Complexity
+Artifacts are primarily dictated by a single effect with the occasional mix-up of having permanent effects (many of which are underwhelming).
+Activation stimuli are similar: the only sliding scale to adjust with how difficult something is to activate is just how hard it is to do that singular trigger.
+
+Since these triggers are always placed in isolation, unless the effect is having some pronounced impact on the crew's ability to trigger the artifact, triggers mostly devolve into incredibly simple and routine tasks.
+
+A water trigger should have lots of depth, but it instead is mostly just walking in with a cup of water and splashing it, which is really the most boring way to engage with something like that.
+
+### Lack of Progression
+Artifacts have an innate progression in the form of the scaling of nodes, which is mostly built around increasingly difficult triggers and more dangerous and wacky effects.
+This is a good start for a system like this, but the unfortunate reality of it is that the scaling isn't pronounced enough to often feel like a deviation from the early-depth nodes.
+
+Especially when taken in with randomization of artifacts, you can oftentimes just get subpar generation that flounders in weak effects that don't feel like a progression in research.
+
+## Artifacts
+The primary changes to generation are fairly simple. 
+While all the nodes still generate in an interconnected structure, they will have more connections to lower depths, resembling a tech web more than a basic tree.
+
+All nodes can have one of 3 basic states:
+- Active
+- Latent
+- Locked
+
+**Locked** nodes are the basic state of all nodes on an artifact.
+All nodes start as locked and they have no effects.
+The primary goal is to unlock these nodes.
+
+Once all of the necessary triggers are activated (will be elaborated on later), a locked node can become a **latent** node.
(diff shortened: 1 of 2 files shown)


Partmedia:
Thanks for your proposal. I recall discussing some of the drawbacks of the current system with you at some point and am happy to see that you're working on some new mechanics.

I generally agree with the problems that you identified with the current system in your "Background" section.

## Artifacts
I had some trouble understanding the "Artifacts" section. I wish the new design proposal linked the original or re-explained a bit of the jargon. I've been around for a little while but I still had difficulty recalling what a "node" is and exactly how "triggering" is different from "activating" a node.

> This simply does all the effects of the active nodes all at once

But as far as I understand, an artifact can only be in one node at a time? Or does it include all previously "activated" nodes?

> with over-use of direct activation potentially "breaking" nodes

So is there "indirect" activation?

Overall I felt too lost in the jargon to understand how the proposal is supposed to work.

I feel like a more concrete example of an artifact, its "nodes", "active effects", and "triggers" would help me better understand what's going on.

## Artifact Equipment

> The artifact analyzer and analysis console gets the most immediate and useful changes. It will now visually show all nodes in the correct structure

Instead of "correct," do you mean "whole?" If the analyzer now shows the whole node structure so that science players can make informed decisions about what paths to take, then I do think it would be a plus for "more player agency".

## Artifact Fragments

Splicing sounds like an interesting and neat way to combine effects.

## New Equipment

No complains. Though, if artifact durability is a mechanic to prevent repeatedly spamming artifact effects, then artifact glue probably needs some balancing to make sure it doesn't re-introduce effect spamming too much.

## Artifact Reports

It's not clear to me if this is like a research report, where if you've blown up your actual artifact, you can still publish on basis of what you've written down in the report. If so, how do you prevent players from writing multiple reports about the same artifact and submitting all of them? (This is solved in real life by reviewers rejecting your paper if it doesn't have enough "novelty")

I hate to be the ideas guy here, but what if the part of the process was made a bit more interactive, say, by the player filling out a multiple choice questionnaire in lieu of automatically generating a report? e.g. "Question 1: For this artifact in its first state/node, did it respond to spilling water? Y/N" This would reward folks for taking notes/remembering and generally being good actual scientists, and still achieve the main objective of letting people generate research points even if the artifact gets destroyed.

## Conclusion

I would approve this design doc on the basis that it sounds like an improvement compared to what we have now. If you would like to spend some time fleshing this out:

- I'm still having a hard time understanding how artifacts relate to "nodes" and "triggers" and "effects." I feel that there's a bit too much abstraction and jargon for me to understand what's going on. Perhaps a better way to think about this is, if you were writing a guidebook entry about XenoArch Redux after this was all implemented, how would it read? I'm more interested in the player interface rather than if behind the scenes it's represented by a graph.
- I like the new proposed equipment and fragments.

EmoGarbage404:
Thanks for the review. I have some general clarifications and musings. if you think they're good, i'll add them into the existing doc.

# Artifacts

> I had some trouble understanding the "Artifacts" section. I wish the new design proposal linked the original or re-explained a bit of the jargon. 

I can elaborate a bit more on the current design just to clarify. As it is now, artifacts consist of interconnected nodes, each one carrying an effect and a trigger. The effect is just some crazy behavior that happens in response to the trigger, which is just some kind of generic action taken upon the artifact.

These nodes form a tree, wherein each individual node's depth within this tree determines the craziness of the its effect and trigger.

> I've been around for a little while but I still had difficulty recalling what a "node" is and exactly how "triggering" is different from "activating" a node.

In the current system, triggering and activating a node are one and the same. It's the new system which makes this distinction, where an artifact as a whole is activated and nodes being triggered is what unlocks new behaviors for activation.

> But as far as I understand, an artifact can only be in one node at a time? Or does it include all previously "activated" nodes?

This is another deviation from the current system. Currently, the artifact is always set to one node in its tree. This is what determines what effect and trigger is present at a given time.

> So is there "indirect" activation?

Lol. The clarification of the 'direct' activation is just to specify the new system where the artifact can be activated in hand to produce various effects. This direct activation is what's referenced later in relation to the glue. This is different than the effect that happens when a node is unlocked, which is just a single effect happening, instead of the activation wherein multiple occur at once.

## Clarification^2

> I feel like a more concrete example of an artifact, its "nodes", "active effects", and "triggers" would help me better understand what's going on.

I'll use the element of a [tech tree](https://en.wikipedia.org/wiki/Technology_tree) as a reference.

Each node is essentially an upgrade in a tech tree. They all connect in the conventional structure that one would expect (a [directed acyclic graph](https://en.wikipedia.org/wiki/Directed_acyclic_graph)).

Just like the current system, all of these nodes have a trigger and an effect associated with them. However, unlike the current system, you do not 'move' to a node, but you instead unlock it like a tech tree.

The 'cost' for unlocking is the activation of all of the triggers of the nodes in that path--that is, all of the nodes that needed to be unlock in order for the current node to become available. 

In this situation, the 'active' nodes (specified in the doc as those that exhibit effects when the artifact is activated) are the nodes in each path that have the highest tier. The remaining nodes are those that are classified as 'latent'--unlocked but not exhibiting behavior.

# Artifact Equipment

> Instead of "correct," do you mean "whole?" 

Yes.

# New Equipment
My one comment about the artifact glue concerns is that artifact fragments are only obtainable from salvage or from destroying existing artifacts using an unlock. This means the supply is inherently quite limited, and with artifexium being a useful utility already, artifact glue is probably going to be rare enough to not be a massive balance concern.

# Artifact Reports
This is the only section that I am very much okay with dropping. If you don't think it adds meaningful complexity, I'm totally okay with dropping it wholesale.

The current system works with a 'novelty' system, where the research generated by the artifact is stored, and the new amount must exceed that stored value in order to generate research.

> if you've blown up your actual artifact, you can still publish on basis of what you've written down in the report.

That is the intent, yes. You would be able to essentially save a 'snapshot' of the artifact a given time and any snapshot could be submitted for the report.

> If so, how do you prevent players from writing multiple reports about the same artifact and submitting all of them? 

The proposal outlines that the new report must exceed the original one in the calculated value, at which point you earn the difference, which is essentially the current system. I must admit that this is basically just the old system but with added complexity, and I'm not particularly loving it on this second inspection.

> I hate to be the ideas guy here [...]

I think making it more of an interactive form could potentially be interesting, especially if only a few reports are being created per artifact. However, my re-examination leads me to think there isn't much value to that without more drastic changes.

EmoGarbage404:
@Partmedia just pushed some changes including a pretty big rework of the research generation section as well as better clarifications throughout the doc. Feel free to merge at this point as i think it's pretty much gtg.

Partmedia:
Good luck!

src/en/proposals/emogarbage-xenoarch-redux.md
```diff
@@ -0,0 +1,173 @@
...
+- Missing nodes, such as those from the effect of the Arti-nUKer.
+- Additional nodes, such as from the effects of the Splicer. 
+
+Note that the calculation for the last two points is based on the total number of nodes in the original vs. the current artifact.
+If you destroy 2 nodes and then repair it by splicing 2 nodes onto it, you incur 0 penalty.
+
+To actually gain research from the artifact, you must place it on an analyzer and begin the 'research' task in the analysis console UI.
+This begins a 30 second countdown where the artifact must remain on the analyzer, cannot be activated, cannot have any stimuli trigger, and the analyzer must remain powered.
+
+This provides an interesting window wherein sabotage and other such measures can be taken to steal the artifact or otherwise interrupt science.
```
deltanedas, line 171: late but this is just called power flickering, which will get very annoying

PR #209, opened 2024-04-27
Chaplain Diversification

Terraspark4941:
See markdown

### DIFF
--- a/src/en/space-station-14/departments/service/proposals/chaplain-diversification.md
+++ b/src/en/space-station-14/departments/service/proposals/chaplain-diversification.md
@@ -3,15 +3,15 @@ AKA. The Quran update
 
 | Designers | Implemented | GitHub Links |
 |---|---|---|
-| Terraspark4941 | :x: No | TBD |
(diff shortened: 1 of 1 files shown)


Terraspark4941:
okay something's up with the SUMMARY.src whitespace someone please help

slarticodefast:
Hey, thanks for the proposal.
Since you made this PR a few things had changed in the content repository, in particular multiple bibles were added and then partially removed again. This also had some discussion in the maintainer meeting.
See this PR https://github.com/space-wizards/space-station-14/pull/39698

To summarize:
- We would like to avoid real life religions, so we made most bibles fictitious. The bible was renamed to "space bible" to move it a little further away from the real thing (but it's not really possible to completely unchristianify the chaplain).
- Different bibles are now found in the character loadout instead of using a radial menu.
- Different effects for each bible would be up for discussion, but they would need to be somewhat balanced so they do not become a powergaming choice despite the chaplain being mostly a roleplaying job.

Otherwise I think these changes are minor enough to not need a design doc, so I will be closing this PR.

PR #213, opened 2024-05-02
Grapple System

Mith-randalf:
There's enough words in the doc I dont think I should add more here. Here's what I think a good grapple system would be, very heavily based on SS13 mechanics across a few codebases.


Mith-randalf:
on consideration the mutual pin thing is dumb and I'm gonna rewrite this to remove it later

mirrorcult (approved):

mirrorcult:
this is genuinely really really good, more or less what i've wanted for grapple combat but with more substance and some better ideas. i think people could dislike the random elements but imo its great for something like this, im happy to merge in its current state unless you want to make other changes

Mith-randalf:
Neat. Yeah I do wanna make some changes, it's been percolating in my head the last few weeks and I've got a couple tweaks I'll get done over the next few days. Nothing huge.

Mith-randalf:
Okay those are the changes I wanted to make; I removed 'mutual pins make you tumble in a rolling grapple' because prediction over two fighting inputs seems bad to implement and I think mutual Grabs needed a more straightforward win-state, and I added pulling a mob applying a small slowdown which can be stacked by multiple mobs pulling, because I think 'getting dogpiled by 6 zombies' should be harder to walk away from than it is.

PR #214, opened 2024-05-03
Reduced Metagaming Mechanics

SlamBamActionman:
This proposal aims to reduce the impact metaknowledge has on the game by shifting it from being a rules-enforced system to a mechanics-enforced system. This will reduce ambiguity for what constitutes a rulebreak, reduce admin workload and better align player knowledge with character knowledge.

The proposal identifies 5 "problems" currently in the game where knowledge about the game is restricted via rules enforcement, and suggests changes that allow players to possess and act upon the knowledge while making it harder to do so without proper justification.

The important part of the proposal is the identification of issues and attempts to solve them; the particulars of a specific solution (e.g. the damage proposed in the "Implant Problem") is malleable based on feedback.

### DIFF
--- a/src/en/space-station-14/departments/security/proposals/reduced-metagaming.md
+++ b/src/en/space-station-14/departments/security/proposals/reduced-metagaming.md
@@ -45,8 +45,8 @@ Strangely, implant searches are much more restrictive on MRP. To check, Security
 ### Proposed Solution
 
 Security should not have any rule limitations on checking implants, but instead Security should be mechanically discouraged from performing random implant checks. I propose the following:
-- Security has knowledge of implants; that they exist, and what types the Syndicate have access to.
-- Security has no limitations on performing implant checks.
+- Rule change: Security has knowledge of implants; that they exist, and what types the Syndicate have access to.
+- Rule change: Security has no limitations on performing implant checks.
 - Checks can only be performed with the `Implant Extractor`, a device that can be printed in the Medfab.
 - To extract, the Implant Extractor must be set to the specific type of implant to be extracted. A doafter is then performed.
    - If successful, the chosen implant gets extracted into the Implant Extractor similar to how an Implanter is done now.
@@ -66,17 +66,46 @@ Uplinks are in a kind of weird situation where it's one of the most powerful too
 - Security finds an open uplink in the PDA.
 - Security has previously found an open uplink in another PDA, and can thus pre-emptively confiscate any PDA they have reasonable suspicion of having an uplink.
 
-Right now the rules are the main way of preventing overzealous Security confiscating PDAs. Traitors do have a gameplay option as well: The Uplink Implant. This implant does have the benefit of making PDA confiscating a non-issue, but the tradeoff is 2 less TC and with the stringent rules on PDA confiscating it's unlikely to get any benefit from it. There's no reason to keep the uplink open after the implant is used, so you only get value if another uplink has been found. Additionally it suffers from the vulnerability of implant checks.
+Right now the rules are the main way of preventing overzealous Security confiscating PDAs. Traitors do have a gameplay option as well: The Uplink Implant. This implant does have the benefit of making PDA confiscating a non-issue, but the tradeoff is 2 less TC and an implanter that must be disposed of, and with the stringent rules on PDA confiscating it's unlikely to get any benefit from it. There's no reason to keep the uplink open after the implant is used, so you only get value if another uplink has been found. Additionally it suffers from the vulnerability of implant checks.
 
 ### Proposed Solution
 
-The PDA is very easily the single point of failure with this problem, but it's also unique in its ringtone lock. 
+The PDA is very easily the single point of failure with this problem, but it's also unique in its ringtone lock. Confiscating a PDA does come with some cost to Security in that they need to secure a replacement; worth it for a confirmed Syndie, but not viable for all arrests. If we were to remove the limitations on Security randomly confiscating PDAs there would need to be an equally effective method of eluding detection. It would also need to incur a cost for Security to confiscate. I propose the following:
+
+- Rule change: Security has knowledge PDAs can contain Uplinks, and may confiscate them if they have a reason to believe it could contain one.
+- Uplink Implanter cost is reduced to 1 TC, and changes proposed in "The Implant Problem" are also included.
+- The Uplink Patch is a new device that can be bought from the Uplink for 1 TC.
+  - The Uplink Patch can be attached to any object and turns it into an Uplink.
+  - The Uplink can be locked and unlocked by saying one of the agent codewords while wearing/holding/carrying the object.
+  - TC must be inserted into the object to purchase anything.
+
+The Uplink Patch would turn any object into a possible Uplink, making the permutations for what to confiscate anything in the Syndie's inventory. Even if a SecOff would be overzealous and try to take everything, the uplink could be an implant as well. With the changes in the Implant Problem, this would require Security to have strong suspicion of where the uplink would be to actually catch it.
 
 ## The Thief Gloves Problem
 
+Thief gloves come in two varieties, each with their own problem. 
+
+1. Normal Thief Gloves are indistuingishable from black gloves. The only way to know they have been used is if someone finds out an item on their person has been stolen, and even then Security are not allowed to know they must search for black gloves unless an uplink has been previously found. They also stand out for crew who don't normally wear black gloves, making overzealous SecOffs search and possibly even confiscate them.
+
+2. Chameleon Thief Gloves mean to remedy this. However, as they leave holographic fibers on everything they touch it becomes very obvious whenever they have been used. It leads to a rather opaque and unintuitive game of trying to not wear the thief gloves except in the most critical of moments, making finding a spare pair of gloves in someone's bag highly suspicious. They're also trivial to check since it's just to put them on and check whether they allow Chameleon properties.
+
 ### Proposed Solution
 
-## Disguised Items Problem
+- Thieving Gloves and Chameleon Thieving Gloves are merged into a single item: The Thief Glovebox.
+  - When purchased/obtained, the user is given a box. Opening this box activates a UI similar to a Chameleon UI.
+  - Once a glove design is chosen, the user is given a pair of thieving gloves with that design.
+  - The gloves give off matching fibers. They are not Chameleon.
+- Non-engineering insulated gloves/budget insulated gloves now come in 3 different colors with corresponding colored fibers. They are still called "Insulated Gloves"/"Budget Insulated Gloves". 
+- Rule change: Security has knowledge of Thieving Gloves and may confiscate the gloves of anyone who has stolen or attempted to steal contraband.
+- Rule change: An additional rule is added: Security may *not* test Thieving Gloves by putting them on and trying to steal an item off someone.
+
+This means that a thief is safe from having their gloves lost before committing a crime. Detectives can no longer use holofibers to confirm a Thief antag, but must instead rely on the fibers of the gloves around the crime scene. The addition of additional insuls colors means selecting yellow insulated gloves is no longer a safe option to blend in with all the other yellow gloves.
+
+## The Thief Sets Problem
+
+### Proposed Solution
+
+## The Disguised Items Problem
 
 ### Proposed Solution
 
--- a/src/en/space-station-14/departments/security/proposals/reduced-metagaming.md
+++ b/src/en/space-station-14/departments/security/proposals/reduced-metagaming.md
@@ -6,7 +6,7 @@
 
 ## Overview
 
-This proposal contains a list of design changes and features aiming to reduce the impact of metaknowledge in the game. 
+This proposal contains a list of design changes and features aiming to reduce the impact of metaknowledge in the game. While the proposal touches on several different issues the changes proposed affect each other and overlap, which is why they're all compiled here. The overarching theme is moving away from rule enforcement to mechanical enforcement, allowing crew players to possess general knowledge about antags but make it harder to apply it in a round without proper reasoning.
 
 ## Background
 
--- a/src/en/space-station-14/departments/security/proposals/reduced-metagaming.md
+++ b/src/en/space-station-14/departments/security/proposals/reduced-metagaming.md
@@ -18,10 +18,32 @@ The negative impact enforcement of metaknowledge is the following:
 - Players need to self-assess what counts as abusing metaknowledge, possibly over-correcting as there's no ingame feedback.
 - Disagreements about whether someone is abusing metaknowledge (e.g. a SecOff testing a pair of gloves for chameleon) can easily occur between players in a round, and is highly contextual.
 - Admins need to put in time and resources enforcing these rules, which is always desirable to reduce.
+- Players have to pretend they don't know about certain mechanics.
 
 To find out where these problems lie, [I created a thread on the Discord](https://discord.com/channels/310555209753690112/1227226189358174209) to gather feedback from the playerbase where they have percieved metaknowledge abuse. This feedback is used as a basis for the changes and will be cited throughout the proposal.
 
 The proposal will contain both mechanical design changes as well as rule changes. The rule changes are necessary as the goal of the PR is to reduce admin workload, and where it's not possible to eliminate them the goal will be to make them more clear and reduce grayzones. 
 
-## The PDA / Uplink problem
+### Security & Resources
 
+Other than explicit rules and Space Law, the main thing holding Security back is **time**, **materials**, **personnel** and **information**. Performing a search takes time, which holds up a SecOff that then becomes unable to respond or assist in other duties. Needing to utilize guns, stun batons and flashes means a SecOff must return to Sec to refill, and if materials are scarce this wait can be even longer. Similarly, health is a resource which will cause time and personnel loss from the SecOff needing to be treated.
+
+Information helps mitigate these losses; knowing who to search, what materials to prepare and where to go. The more Security has of each resource the more effective it will be, and overabundance of one resource can make up for the lack of another. Some of these resources put strain on other departments however; materials may be necessary to craft or refill vital Security equipment, but those materials come from Cargo's work. Any rule change that relaxes the requirements on Security should instead impose a drain on its resources. For example, guessing uplink codes is usually too much of a timesink and takes up an officer's attention despite it being allowed by the rules, so it's rarely done.
+
+## The Implant Problem
+
+## The PDA / Uplink Problem
+
+```admonish quote
+Confiscating PDAs of anyone you suspect of being a traitor without proof.
+```
+
+Uplinks are in a kind of weird situation where it's one of the most powerful tools for a Traitor and in-universe a very well-hidden feature, but at the same time it's also one of the easiest metaknowledge items in the game. As soon as a player has been found using a traitor-exclusive item Security players have to pretend that the PDA isn't a high target to confiscate. The reason for this is because you generally want antags to be able to use all their TC during a round to make it interesting. As such, the PDA may only be confiscated for two reasons:
+- Security finds an open uplink in the PDA.
+- Security has previously found an open uplink in another PDA, and can thus pre-emptively confiscate any PDA they have reasonable suspicion of having an uplink.
+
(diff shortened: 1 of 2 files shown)


src/en/space-station-14/departments/security/proposals/reduced-metagaming.md
```diff
@@ -0,0 +1,234 @@
...
+
+### Proposed Solution
+
+The PDA is very easily the single point of failure with this problem, but it's also unique in its ringtone lock. Confiscating a PDA does come with some cost to Security in that they need to secure a replacement; worth it for a confirmed Syndie, but not viable for all arrests. If we were to remove the limitations on Security randomly confiscating PDAs there would need to be an equally effective method of eluding detection. It would also need to incur a cost for Security to confiscate. I propose the following:
+
+- Rule change: Security has knowledge PDAs can contain Uplinks, and may confiscate them if they have a reason to believe it could contain one.
+- Uplink Implanter cost is reduced to 1 TC, and changes proposed in "The Implant Problem" are also included.
```
moonheart08, line 76: So, i think this completely breaks the original design intent behind things like the implanter, which was to be a way to evade losing your uplink *at cost of some of your TC.* Alongside that, I think the original design intent was made on sound grounds, with the idea being that the harder it is to fish out your uplink, the more TC you have to pay upfront to keep it.

Here's what I think might be a better solution:
- We introduce the ability to choose the type of uplink you want sometime around role application, with different uplink types having different TC costs.
- We keep PDAs as-is, but **allow you to pay a small amount of TC to transfer it to a new PDA** (likely by simply entering the dialtone into the new one, they're unique enough this should be fine.)
  - This makes keeping your original PDA highly desirable, as you would lose, say, 3-5TC, as a cost for setting up the new uplink, but *not* mandatory.
- We don't adjust uplink injector cost, it really needs to have a cost to it otherwise there's no reason to ever *not* use it, which makes the PDA uplink completely redundant.
- The uplink patch is a cool idea too, I don't think it should be discarded as a concept, but it also shouldn't be basically free (i.e. 1TC) to evade security that hard.
  - The uplink patch as described is decently powerful when used on stuff in your inventory, but where it's way more powerful is when used on random objects in maints or similar, allowing you to make it *impossible* to track down your uplink because it's actually a traffic cone in a locker somewhere. Essentially, it's way more powerful as a concept than this doc gives it credit for.

src/en/space-station-14/departments/security/proposals/reduced-metagaming.md
```diff
@@ -0,0 +1,234 @@
...
+  - When purchased/obtained, the user is given a box. Opening this box activates a UI similar to a Chameleon UI.
+  - Once a glove design is chosen, the user is given a pair of thieving gloves with that design.
+  - The gloves give off matching fibers. They are not Chameleon.
+- Non-engineering insulated gloves/budget insulated gloves now come in 3 different colors with corresponding colored fibers. They are still called "Insulated Gloves"/"Budget Insulated Gloves". 
+- Rule change: Security has knowledge of Thieving Gloves and may confiscate the gloves of anyone who has stolen or attempted to steal contraband.
+- Rule change: An additional rule is added: Security may *not* test Thieving Gloves by putting them on and trying to steal an item off someone.
+
+This means that a thief is safe from having their gloves lost before committing a crime. Detectives can no longer use holofibers to confirm a Thief antag, but must instead rely on the fibers of the gloves around the crime scene. The addition of additional insuls colors means selecting yellow insulated gloves is no longer a safe option to blend in with all the other yellow gloves.
```
moonheart08: I genuinely just don't like this, so this is much more of a personal opinion than a design opinion, but never-the-less here's some opinions for you:
- I think the real issue is that glove colors other than yellow are extremely rare, with little purpose. Some SS13 servers have solved this by making the other gloves actually do things (i.e. sterile gloves are sterile, normal black gloves are *heat resistant*, etc.)
  - Implementing the black gloves heat resistant mechanic and removing the ability for other gloves to do things like pull lights would be a good first step in normalizing them.
  - Thieves gloves are intended to be a stealth mechanic, end of the day, so it is *imo* better to address the things that make them not stealthy instead of making them even stealthier.
- Chameleon gloves work fine as is, especially if the above proposed changes are made to make having multiple glove pairs more common. I don't think we need multiple insul colors to remedy this, just reason to use multiple gloves for different purposes.

src/en/space-station-14/departments/security/proposals/reduced-metagaming.md
```diff
@@ -0,0 +1,234 @@
...
+- Syndie Set
+  - Contains Master At Arms hat and Interdyne Cigarettes. These could betray the Agent ID and Emag.
+- Sleeper Set
+  - Contains a Pyjama hat and Nitrous Oxide tank. These could betray the Hypopen. The jumpsuit isn't as big of an issue since it can be cut into cloth.
+- Communicator Set
+  - Contains a briefcase filled with a thief figurine, neck gater, tacticool jumpsuit and jensen coat. These could betray the Master Comms Key, CyberPen and Voicemask.
+
+### Proposed Solution
+
+Cull the unnecessary fluff items from the set and make them the same level of discoverability. If it's desirable to have the thief hide stuff, include that in the items being recieved; implants require the thief to hide the implanter for example. If the Communicator Set should require the thief to hide a fluff thief figurine for example, instead make the the Master Comms Key come in a box that needs to be hid.
```
moonheart08, line 137: I think there is, once again, a simpler solution here, make the supposedly stealthy items actually stealthy. This includes things like making briefcases good extra storage and making more roles spawn with them (Lawyer, librarian, and other "classy business" service roles especially should have them if they don't already)

Additionally, is it *really an issue* that these items can be in theory found? A number of these are standard station items one could find in, say, maint, and blend in perfectly there if you don't want them. Many of them already pass the sniff test as things you can see under normal circumstances (though the omega soap should probably get swapped out.)

moonheart08 (changes requested):
partial review

moonheart08:
*imo* a lot of the problems laid out in this doc are "this stealth mechanic fails to be stealthy" and tries to solve them by buffing the mechanic. A lot of them could be solved instead by normalizing the things those stealth items pretend to be, like as noted with the thieving gloves/black gloves.

src/en/space-station-14/departments/security/proposals/reduced-metagaming.md
```diff
@@ -0,0 +1,234 @@
...
+
+### Proposed Solution
+
+This is probably the most comprehensive change and touches on things outside the stated problem above, in the interest of making a more streamlined system and help onboarding new players into contraband gameplay. My proposal is the following:
+
+- Rule change: Crew have knowledge about stealth items.
+- *Non-stealth* items can now be determined to be contraband upon being examined.
+  - Improvised contraband (improvised weapons, shivs, makeshift cuffs etc.) have the following text added: "This item is minor contraband."
+  - Items restricted to departments (insulated gloves, RCDs, scalpels, Security gear etc.) have the following text added: "This item is departmentally restricted."
+  - Command items (HoS' secret orders, Captain's Sabre, Nuke Disk, CE's boots etc.) have the following text added: "This item is restricted to Command."
```
moonheart08, line 181: I disagree with this once again on opinion (so it's up to you whether or not you agree)
We shouldn't completely eliminate the benefit of having game knowledge by spelling everything out, especially not at a glance like this such that a player would likely ask to examine the item to figure out if it's contraband instead of remembering the name or look or etc. (in other words, this is a variation of the 'minimap problem' of giving the player too much information making them not engage with parts of the game)

If anything, *itemize these checks*, rather literally. Give security a fancy scanning gun or something with a short doafter for example, encourage players to eventually remember these classifications. I'm not against including these kinds of classifications into the game, but I don't think we should shove them into examines.

src/en/space-station-14/departments/security/proposals/reduced-metagaming.md
```diff
@@ -0,0 +1,234 @@
...
+  - Syndicate items have the following text added: "This item is Syndicate contraband."
+- Note that the change above does not change the crime definitions in Space Law.
+- Most stealth items can now be locked to hide their functionality, making it act as a normal item. If the wearer/holder speaks a Traitor codeword, the functionality unlocks.
+  - While locked, the contraband examine text is not shown. 
+  - Thieves get a single codeword, not shared by other thieves or traitors.
+  - Nukies get a set codeword "Syndicate".
+  - Note that this is not applicable for all items. For example, the Extra-Bright Lantern can not be locked, allowing it to only pass inspection at a glance.
```
moonheart08, line 188: This is a good idea :)

(i figure I should at least note approval of some of these things after railing so hard on it)

src/en/space-station-14/departments/security/proposals/reduced-metagaming.md
```diff
@@ -0,0 +1,234 @@
...
+- Security has a new machine available roundstart, the **Contraband Scanner**.
+ - The device acts like an airport x-ray scanner, able to find hidden compartments and technologies in otherwise mundane items. 
+  - An item can be placed inside the scanner. After six minutes the item comes out; if the item is of Syndicate origin, it is marked as such when examined: "This item is disguised Syndicate contraband".
+  - This mark can be cleaned off with soap.
```
moonheart08, line 192: This is also a good idea, and lines up with what i mentioned earlier with my opinion on the examine text (I hadn't read this bit yet.)

src/en/space-station-14/departments/security/proposals/reduced-metagaming.md
```diff
@@ -0,0 +1,234 @@
...
+```admonish quote
+"Quiet shift today"
+```
+```admonish quote
+"I found a cobra or some other syndicate tech on station/a syndicate has done something noticable. Main round antag is traitors."
+```
+```admonish quote
+"It's been strangely quiet and nothing has really happened in the first 30 minutes. Probably nukies"
+```
+Can be solved by adding single Traitors to non-Traitor gamemodes: https://github.com/space-wizards/space-station-14/pull/27501
```
moonheart08, line 215: Can also be solved by proper implementation of "dynamic" modes, adjusting map loot to also contain more "fun" or incriminating items (maybe even faking DNA on items to show someone might have used them before the round took place?), etc.

moonheart08 (changes requested):

src/en/space-station-14/departments/security/proposals/reduced-metagaming.md
SlamBamActionman: Just wanna start by saying I appreciate the comprehensive feedback, really glad to see the topic get engaged!

I think that the main reason my suggestions regarding uplinks are so forgiving for antags is that I've leaned hard into the idea that confiscated TC is unfun. Sec should absolutely be rewarded if they find an uplink and similarly the Traitor should be punished for letting it be found, but if the uplink knowledge rules are relaxed there needs to be a safeguard against a Sec confiscating a PDA "because you had a syndie item and I know uplinks are in PDAs". 

Basically if you allow Security to know where locked uplinks may exist, you have two options. Either make it too difficult for Sec to just guess where the uplink is, or make it not punishing if Security takes your locked uplink. I really like your idea of being able to transfer the uplink to a new object, but I would really reduce the cost of it to 1 or 2 TC. That way you get the full punishment if you leave your PDA unlocked, but if Sec decides to be ornery and confiscate your locked PDA you aren't harshly punished for it.

You're right that the uplink patch does have more power than I gave it credit. However I would still keep it at a low TC. The reason is the same as what I give for the implant; it's highly situational, doesn't offer anything on its own and especially if you hide it in maints it's vulnerable to being grabbed by some random person. If Sec can know about how uplink works then Traitors must be given options for cheap to counteract it.

src/en/space-station-14/departments/security/proposals/reduced-metagaming.md
SlamBamActionman: I think that your approach is fully valid, just maybe harder to balance for as there's a risk one kind of gloves is always the "meta". 

The reason why I'm for removing the Chameleon Thief gloves (and its corresponding fibers) is because if Security are allowed knowledge of how Thieving Gloves work, it becomes trivial to know they are in play the moment the victim says "My X was stolen off me". However making the gloves a single type (like the normal gloves) have its own issues as stated in the proposal, so the glovebox is a middle ground.

Ultimately the goal is glove diversity. Whether that's achieved via making more gloves viable, or making the viable gloves more varied, either works for me. :)

src/en/space-station-14/departments/security/proposals/reduced-metagaming.md
moonheart08: I'm going to be very firm on the patch thing, it's *really not hard in my experience* to make it extremely unlikely someone will ever find an item even on an 80 player station, especially if secret rooms/etc are used.

This already technically makes hiding your uplink really easy if you're smart about it, and I've used tactics like that before to terrorize sec as a greytider because they can't find any of the junk I took from them. (Turning yourself in after hiding all your stuff in a room nobody ever visits is an extremely effective tactic.)

So apply that knowledge to a random, seemingly worthless item (say, a cig butt), and put it in a infrequently used room in maintenance somewhere. Nobody will ever find it, and you paid effectively nothing for that privilege.

Also, a thing I actually forgot to mention, most SS13 servers have *two* free uplinks you can choose from, the standard PDA uplink and the less common, but still cost free headset-based uplink. It'd likely be good to implement the headset-based uplink.

src/en/space-station-14/departments/security/proposals/reduced-metagaming.md
moonheart08: I will note, I think the thieving gloves themselves are a mis-executed stealth item, and I'm very unsure on how to actually fix them because their specific usecase (theft without an indicator) is too specific to actually *be* a good stealth item.

src/en/space-station-14/departments/security/proposals/reduced-metagaming.md
SlamBamActionman: Some items I agree pass the sniff test, such as the Syndicate blanket or Interdyne cigarette pack. Seeing one of those does not *immediately* raise alarm. But a thief figurine is basically a blaring red light that there is a thief on the station, especially since that kit specifically has a bunch of fluff that largely gets tossed away in the same pile.

It's stupid that I can find a vodka bottle in the medbay backroom and immediately think "A doctor has a storage implant". It's just pointless. One can probably blame it as skill issue on the Thief's part, but unlike an implanter which is intrinsically tied to an item acting on a vodka bottle just feels cheap.

src/en/space-station-14/departments/security/proposals/reduced-metagaming.md
SlamBamActionman: On one hand I agree that this does lower the bar on a certain mechanic, i.e. knowing what is and isn't contraband. On the other hand that's also a mechanic largely enforced via rules exclusively found on an external website outside the game, which isn't the best player experience for a new Sec Cadet. 

This system also acts both ways. It allows non-Sec crew to be aware of what constitutes contraband even if they haven't played Sec or engaged with that system. Which is nice to new players, and also solves the pet peeve of that one tider who refuses to give up their baseball bat because "it's a sports tool, it's not contraband weeeh" so you need to quote the whole contraband list to them.

src/en/space-station-14/departments/security/proposals/reduced-metagaming.md
moonheart08: > Some items I agree pass the sniff test, such as the Syndicate blanket or Interdyne cigarette pack. Seeing one of those does not _immediately_ raise alarm. But a thief figurine is basically a blaring red light that there is a thief on the station, especially since that kit specifically has a bunch of fluff that largely gets tossed away in the same pile.
> 
> It's stupid that I can find a vodka bottle in the medbay backroom and immediately think "A doctor has a storage implant". It's just pointless. One can probably blame it as skill issue on the Thief's part, but unlike an implanter which is intrinsically tied to an item acting on a vodka bottle just feels cheap.

Is the thief figurine not standard maint loot? (I'd consider that a bug)

Any items that aren't standard maint loot should either be added to that pool or removed from thief kits, that I can agree on.

src/en/space-station-14/departments/security/proposals/reduced-metagaming.md
moonheart08: > On one hand I agree that this does lower the bar on a certain mechanic, i.e. knowing what is and isn't contraband. On the other hand that's also a mechanic largely enforced via rules exclusively found on an external website outside the game, which isn't the best player experience for a new Sec Cadet.
> 
> This system also acts both ways. It allows non-Sec crew to be aware of what constitutes contraband even if they haven't played Sec or engaged with that system. Which is nice to new players, and also solves the pet peeve of that one tider who refuses to give up their baseball bat because "it's a sports tool, it's not contraband weeeh" so you need to quote the whole contraband list to them.

Here's an idea: make this a PDA app (NT Certified Object Identification Cloud Service or something else suitably silly), we can offload other "extraneous" information about items into it alongside contraband status in the future if we come up with further uses.

src/en/space-station-14/departments/security/proposals/reduced-metagaming.md
SlamBamActionman: I did consider a headset uplink and had it written out in an initial draft, but found that it bumped up against the issue of an overzealous Sec confiscating *both* the PDA and headset. I'm not opposed to it (I think it'd be a cool feature!) but because of that I rolled it into the patch suggestion. 

By the sound of things we're getting to the number tweaking part of implementation for the patch. Even if it ends up more expensive than what I consider optimal it still sounds like a good addition to the game.

src/en/space-station-14/departments/security/proposals/reduced-metagaming.md
moonheart08: making this an app would also tie nicely into existing guidebook systems, and it could potentially pull up related topics for an item.

src/en/space-station-14/departments/security/proposals/reduced-metagaming.md
SlamBamActionman: As someone who is a prolific user of the thief gloves as both thief and traitor, the gloves are incredibly effective when used in conjunction with chaos. The stealth usefulness they provide is for the moments before and during the theft; once you've absconded and the victim realizes they've lost their stuff it's important to hide the gloves as they do not stand up to targetted inspection. 

Considering the alternative is that you need to kill your target to steal from them I think they are in a good place when it comes to functionality.

src/en/space-station-14/departments/security/proposals/reduced-metagaming.md
SlamBamActionman: > Here's an idea: make this a **default** PDA app (NT Certified Object Identification Cloud Service or something else suitably silly), we can offload other "extraneous" information about items into it alongside contraband status in the future if we come up with further uses.

Works for me! Afaik the only reason why contraband isn't listed in the guidebook is because... it's not easy to automate? Honestly not sure. Having a comprehensive list in the PDA that looks nice and has a snazzy UI would be pretty neat tbf.

The exact implementation is up in the air but as long as we can mark items in the .ymls with a ContrabandComponent the system can then represent that information in whatever way works best.

src/en/space-station-14/departments/security/proposals/reduced-metagaming.md
SlamBamActionman: Thief figurines *technically* can spawn on the map. But they're only available in a specific spawner meant to be used in Libraries and competes with all the other figurines, so in practice they're exceedingly rare. 
https://github.com/space-wizards/space-station-14/blob/0450a416e22651270b5dc8f270ac5ca6780b7075/Resources/Prototypes/Entities/Markers/Spawners/Random/toy.yml#L136

src/en/space-station-14/departments/security/proposals/reduced-metagaming.md
moonheart08: Automating this is relatively trivial we just don't have a nice system for syncing lists the server invented during startup with the client. (This is something the chemistry guidebook has to have a custom solution for, for example.)

Generalizing that would make it more likely someone would come do it, I might try to do so later.

SlamBamActionman:
Added the Contraband Usage Problem, which came out of discussing the issue of Sec/Command being rules-limited from using Syndicate items.

mirrorcult:
pre empt merging this because im going to be working on some of these

PR #218, opened 2024-05-03
Design labeler all match rule

VasilisThePikachu:
I think this works?????


VasilisThePikachu:
This seems to work. Scream at me if it does not.

PR #221, opened 2024-05-14
Chatfactor Design Page

FairlySadPanda:
This document outlines a lot of information about how Chat works in the "Chatfactor", the rework of backend chat code (that will then lead to a rework of frontend chat code).


src/en/space-station-14/chat.md
```diff
@@ -0,0 +1,637 @@
...
+* Background: Where automated chats like vending machine announcements are issued. Essentially identical to local chat, but these messages are "not real" and do not go into chat logs or get message IDs (see below)
+* Prayer: Messages sent from players in-game to admins via some artifact like the red phone or the maint altar.
+* Subtle: Messages targeted at one specific player, currently solely an adminbus feature.
+
+These chat types cover messages that are "round-scoped". This means they relate to something that shows up in replays and are "things that happened in that particular round".
+
+### "Out of round" messages
+
+* OOC: Lobby comms for people out-of-game. Usually switched off by cvar during games.
+* Admin: Admin notices, from all-player bwoinks to requests for feedback on events.
```
VasilisThePikachu, line 27: Either I'm misreading this, or you may have confused the purpose of admin chat

Admin chat (nicked bubblegum chat due to it being pink and making a popping sound on each message sent) is meant as a chat for admins to talk to each other in game with other admins. Players/deadmined should be unable to see this chat.

VasilisThePikachu (commented):
Due note you will need to add this to the "summery.md" file or be linked to somehow. Otherwise mdbook won't render it

moonheart08:
I'm not a fan of the code I'm seeing presented, it doesn't respect the data driven approach of the game's design and instead has custom copy-pasted handlers for everything. 

Why aren't chat types prototypes with fields dictating behavior? Needing a new event pathway (i.e. AttemptDeadChatEvent and co) for every new channel type is a design and maintenance hazard, especially when they all share common and heavily copy-pasted code (an issue that needs resolved with the *current* chat implementation) instead of sharing a common pathway and only doing things differently as necessary for that form of communication.

moonheart08:
> I'm not a fan of the code I'm seeing presented, it doesn't respect the data driven approach of the game's design and instead has custom copy-pasted handlers for everything.
> 
> Why aren't chat types prototypes with fields dictating behavior? Needing a new event pathway (i.e. AttemptDeadChatEvent and co) for every new channel type is a design and maintenance hazard, especially when they all share common and heavily copy-pasted code (an issue that needs resolved with the _current_ chat implementation) instead of sharing a common pathway and only doing things differently as necessary for that form of communication.

For example, OnAttemptEmote, OnAttemptLocalChat, OnAttemptLooc, OnAttemptEquipmentRadio, OnAttemptInternalRadio, and OnAttemptWhisper are all *identical* to one another and only differ by behavior that could be better expressed as config options (or even, special handler types, give the prototype a field of IChatHandler (or multiple fields, of distinct behaviors) and put implementations configured to that form of communication in them.)

moonheart08:
While this is largely my personal opinion, my personal opinion is we should be *absolutely minimizing* all forms of copy-paste or redundant code in the new system, because it's a major part of what makes the old system so awful to work with.

mirrorcult:
> I'm not a fan of the code I'm seeing presented, it doesn't respect the data driven approach of the game's design and instead has custom copy-pasted handlers for everything.
> 

yeah, this was an issue brought up in the original PR for this and i dont really understand what part of it is necessary for the architectural design changes that are valuable here

src/en/space-station-14/chat.md
FairlySadPanda: Oh, fair. Tbf I've not touched this aspect of the code at all. Will separate admin chat from bwoinks

FairlySadPanda:
> For example, OnAttemptEmote, OnAttemptLocalChat, OnAttemptLooc, OnAttemptEquipmentRadio, OnAttemptInternalRadio, and OnAttemptWhisper are all _identical_ to one another and only differ by behavior that could be better expressed as config options (or even, special handler types, give the prototype a field of IChatHandler (or multiple fields, of distinct behaviors) and put implementations configured to that form of communication in them.)

This is implementation detail suitable for feedback on this PR: https://github.com/space-wizards/space-station-14/pull/27740 as that's where I've drawn the examples from.

FairlySadPanda:
(shoutouts to fat-fingering the close button, yippee)

FairlySadPanda:
I've been focussing on splitting things out into very small, discrete events and components. This is intentional. Each "channel" we currently have has different rules for how it works. For example, whispering differs from emoting. As such, they are different code pathways.

There will be some points where code pathways converge (for example: all chat commands enter and exit the chat repository via the same paths) and some where they diverge. But at no point does a chat message lose its typing as a message of its type. Sometimes (honestly, mostly) the type is reflected by generics.

The key architectural decision is to _rely on the event bus for directing code flow_, not handlers themselves. That's why there's a full set of Attempt messages, not just one attempt message that has variable internal data.

I'd encourage folks to give feedback on 27740 so I can iterate on the specifics of this implementation now, as it's gonna be overwhelmingly the best PR to get this locked down in without other distractions.

Not trying to be a tyrant here: what works for the project is deffo gonna be different to my rabbit hole.

For Mirror: happy to discuss in #contributors whenever. I know Moony doesn't use Discord so if they've got a preferred way of discussing things lmk, as Github is a tad clunky for back-and-forth.

FairlySadPanda:
(if there's a tldr to my architecture here, it's "we need to stop thinking about chat as one big system, and think of it as a set of systems that share common code")

moonheart08:
Going to be 100% frank: The hard typing thing is a terrible idea due to the sheer amount of code duplication it is inflicting on the codebase. You do not need strict custom pathways for every chat type when most of them share the exact same (or very similar) behavior.

Additionally, when you started including explicit code in your doc I think you lost the "implementation detail" card, the implementation is rather barely specified within this document.

I would do the following:
- Completely ditch the distinct chat events, have singular ChatAttemptEvent, ChatValidationEvent, and ChatFailedEvent.
- Have a singular core record type, called ChatMessage.
- Attach some form of composable data storage to ChatMessage, possibly an entity, possibly just a `<Type, Object>` dictionary, and store relevant but not required information within it. This would be things like language information for IC chat
- Attach a prototype to ChatMessage, use its fields to dictate logic and optionally *dispatch typed events from the prototype as needed* from there. <br/>
   There's a simple pattern you can use here, have a base ChatEvent record type, and have nullable fields on the prototype that would contain said type. To dispatch such an event, you do `RaiseLocalEvent((object)(template with { ChatMessage = message })`, allowing C#'s with syntax to figure out how to clone the type. (Note the object cast is necessary to tell the engine to use the non-generic pathway for raising an event, so listeners can listen for their concrete type of choice.)
- Every time you would find it convenient to just copy-paste something, **stop**, step back, and find another approach.

FairlySadPanda:
> "You do not need strict custom pathways for every chat type when most of them share the exact same (or very similar) behavior."

I think this is where the debate is and where opinions will differ most strongly.

There are some technical problems that need to be solved by any solution:

1. There are general and specific rules at each stage. Specific rules are almost always tighly-bound to one particular type of chat.
2. These general and specific rules need to be written and maintained by both upstream and downstream maintainers. By necessity, upstream should avoid making hard-to-change decisions so downstream can modify the rules as they see fit.
3. The code needs to scale horizontally - more types of chat channels - and vertically - more complex chat channels.

We can't really stop downstreams from mangling their code (as much as it'd be nice). We also can't really generate a perfect implementation. The key is to write an implementation that can survive long-term and has patterns that can be grokked by a reasonable developer.

My concern with the weakly-typed approach is that it appeals to "obvious" simplicity - "we don't need bespoke things for similar work" - and just hides the complexity problem. What you're essentially defining with "one big set of chat events with mutable internal structure via prototypes" is essentially generics with extra steps anyway.

moonheart08:
> > "You do not need strict custom pathways for every chat type when most of them share the exact same (or very similar) behavior."
> 
> I think this is where the debate is and where opinions will differ most strongly.
> 
> There are some technical problems that need to be solved by any solution:
> 
>     1. There are general and specific rules at each stage. Specific rules are almost always tighly-bound to one particular type of chat.
> 
>     2. These general and specific rules need to be written and maintained by both upstream and downstream maintainers. By necessity, upstream should avoid making hard-to-change decisions so downstream can modify the rules as they see fit.
> 
>     3. The code needs to scale horizontally - more types of chat channels - and vertically - more complex chat channels.
> 
> 
> We can't really stop downstreams from mangling their code (as much as it'd be nice). We also can't really generate a perfect implementation. The key is to write an implementation that can survive long-term and has patterns that can be grokked by a reasonable developer.
> 
> My concern with the weakly-typed approach is that it appeals to "obvious" simplicity - "we don't need bespoke things for similar work" - and just hides the complexity problem. What you're essentially defining with "one big set of chat events with mutable internal structure via prototypes" is essentially generics with extra steps anyway.

The approach given is *not* weakly typed, if you actually read it. It uses strongly typed pathways for chat specific behaviors as needed without duplicating code everywhere.

Additionally, *I am a downstream developer.* I am asking for an approach that benefits both you in the long term and *me*, as it means I have less shit to maintain *and* can easily swap out behavior by adjusting prototypes to use my own handler events. The approach I gave is optimized for this exact usecase of "there's a highly generic pipeline, but it has specific behaviors it needs augmented with". This is *the* usecase for events-as-hooks, and I described how to implement those events in a strongly typed fashion already.

This codebase operates on data oriented design, and you're thinking only in terms of what OOP can do when it comes to whether or not it's "strongly typed".
It is strongly typed, it has a prototype describing the exact details of it, it's just not using the *language*'s types.

Do not treat C#'s type system as the only way something can be strongly typed, it's not the only type system in this codebase and it's not suitable for the composeability chat actually needs.

FairlySadPanda:
I'm glad to hear you're doing downstream stuff and are thus a customer of this code - it means you've got a valuable perspective on this.

I'd prefer if you can write up a formal proposal for your system so I can get it checked by maints and I can get on with implementation if they're satisfied.

FairlySadPanda:
Guy successfully burned on bonfire, I now have clear direction. Onwards and upwards.

PJB3005:
FWIW I think this PR's design is fine. Copy pasting isn't really a problem in this context. I discussed it with panda in VC and I do believe the code could use some iteration to compress it down and remove the *egregious* copy paste, I think the general approach is fine. I don't think we *need* to overengineer this system so chat channels can be defined entirely in YAML. Radio channels and IC languages sure, but those should just be their own distinct prototypes.

PR #225, opened 2024-05-25
Art Guidelines

EmoGarbage404:
mythical doc ive been sitting on for ages

could do with some pics but i do what i can

### DIFF
--- a/src/en/space-station-14/art.md
+++ b/src/en/space-station-14/art.md
@@ -84,21 +84,24 @@ This gives you a clear view of the top as well as the front of the object.
 * **Mobs**, **clothing**, and **items** are rendered in a flat perspective. 
 A flat perspective can be thought of as looking at the image straight-on at eye level.
 This maximizes detail while minimizing any perspective-based distortion.
+When perspective is necessary for conveying something properly, use 3/4s.
 
 
 * **Floor tiles** and **ground** textures similarly lack any particular perspective.
 The actual floor itself should be a simple birds-eye perspective but any small detailing like stones or pits can employ a more conventional flat perspective.
 
 
-* **Walls** and wall-like structures (windows, grilles, etc.) are rendered in a exaggerated top-down perspective typically referred to as _classic_.
+* **Walls** and wall-like structures (windows, grilles, etc.) are rendered in a exaggerated top-down style typically referred to as _classic_.
 Walls shouldn't have any kind of 3/4 perspective and, while the sides may be technically visible, the focus is on the "top" of the wall.
+This isn't to say you're seeing inside of the wall into some kind of inner lattice structure, but rather it's more of a flat texture.
+Walls are very difficult and strange in their form so it's best to just look at existing walls for inspiration.
 
 
 * Similarly, **wall-mounted structures** are additionally rendered in a flat perspective.
 This is most visible in things like posters and aids in readability.
 
 ### Color
-* Keep color palettes **limited and distinct**.
+* Avoid many overlapping palettes, or _gradients_ of many similar colors, prefer small, distinct palettes.
 Make sure the individual colors in a sprite have enough contrast to be discerned in game.
 Remember, sprites in game are a lot smaller than your editor: err on the side of high-contrast.
 
@@ -157,8 +160,8 @@ Make sure that the outline is colored according to the section it surrounds, ins
 Using multiple colors in your outline gives a much better result.
 
 
-* Make sure your outlines have shading as well.
-The bottom of an object will have a darker outline than the top of it.
+* Don't include shading on your outlines.
+The value of the outline should be consistent across the entire shape.
 
 
 ### Miscellaneous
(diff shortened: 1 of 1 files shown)


src/en/space-station-14/art.md
```diff
@@ -1,5 +1,233 @@
...
+
+### Perspective
+
+* **Structures** in the game world should be in a 3/4 perspective.
+This includes various machines, furniture, and decor.
+A 3/4's perspective means that they are angled straight forward with the viewer looking down on them from above. 
+This gives you a clear view of the top as well as the front of the object.
+
+
+* **Mobs**, **clothing**, and **items** are rendered in a flat perspective.
```
Flareguy, line 84: This section implies that items should never be 3/4ths, which is simply not true. Items like boxes and power cells need the definition to work. There should be some kind of addendum to this mentioning that.

src/en/space-station-14/art.md
```diff
@@ -1,5 +1,233 @@
...
+* **Mobs**, **clothing**, and **items** are rendered in a flat perspective. 
+A flat perspective can be thought of as looking at the image straight-on at eye level.
+This maximizes detail while minimizing any perspective-based distortion.
+
+
+* **Floor tiles** and **ground** textures similarly lack any particular perspective.
+The actual floor itself should be a simple birds-eye perspective but any small detailing like stones or pits can employ a more conventional flat perspective.
+
+
+* **Walls** and wall-like structures (windows, grilles, etc.) are rendered in a exaggerated top-down perspective typically referred to as _classic_.
```
Flareguy, line 93: "Classic" is just the term used to refer to the style of walls compared to other SS13 walls, as far as I know. I don't think "classic" is an actual term for perspective.

src/en/space-station-14/art.md
```diff
@@ -1,5 +1,233 @@
...
+Walls shouldn't have any kind of 3/4 perspective and, while the sides may be technically visible, the focus is on the "top" of the wall.
```
Flareguy, line 94: This section may imply that we want people to make walls look like material walls / ore walls which have visible "innards," which is *DEFINITELY* not the case.
With that being said, walls are extremely hard to do and I'm not really sure how to convey how I think walls should probably usually look. Perhaps it may be best to emphasize that walls don't really have much of a perspective to begin with...? Either that, or just say "top-down" without elaborating.

src/en/space-station-14/art.md
```diff
@@ -1,5 +1,233 @@
...
+
+
+* Similarly, **wall-mounted structures** are additionally rendered in a flat perspective.
+This is most visible in things like posters and aids in readability.
+
+### Color
+* Keep color palettes **limited and distinct**.
```
Flareguy, line 101: While colors *should* be limited and distinct, it's probably important to note that people shouldn't try to min-max their palettes. It's usually more then okay to have a lot of colors in this artstyle.

src/en/space-station-14/art.md
```diff
@@ -1,5 +1,233 @@
...
+
+### Outlines
+* All sprites should have **colored outlines** around them.
+This helps improve readability in the world and clarify the forms.
+Make sure that the outline is colored according to the section it surrounds, instead of being a single flat color.
+* * Outlining an entire sprite in black often looks plain and amateurish.
+Using multiple colors in your outline gives a much better result.
+
+
+* Make sure your outlines have shading as well.
```
Flareguy, line 160: I am pretty sure this goes directly against how most classic sprites are shaded. It may just be best to just remove this point entirely.

Flareguy (changes requested):

src/en/space-station-14/art.md
moonheart08: Heavy agree here, I think a better phrasing and approach is "Avoid many overlapping palettes, or 'gradients' of many similar colors, prefer small, distinct palettes".

EmoGarbage404:
updated with suggested review

src/en/space-station-14/art.md
```diff
@@ -1,5 +1,236 @@
...
+Using especially vibrant colors or baking in glows into the sprite can give extra emphasis to something that _should_ look out of place.
+
+### The Future Never Looked so Old
+* While the setting is unquestionably in the future, the aesthetics of SS14 commonly and routinely reference older technology.
+The station is lit with buzzing incandescent bulbs, people send faxes to one another, and computers are still fat consoles with big glass screens.
+This retro appearance is an important factor of the visual identity of SS14: _embrace it_.
+
+
+* When creating new assets or features, playing on existing real world technology can create an intuitive understanding.
+Instead of depicting some kind of generic 'credit chit,' using paper bills is immediately recognizable.
```
mirrorcult, line 51:
```suggestion
Instead of depicting some kind of generic 'credit chip,' using paper bills is immediately recognizable.
```

mirrorcult (approved):
good doc

Flareguy (approved):
Looks mostly good now, outside of mirror's review.

PR #227, opened 2024-05-29
Inventory Rig Slot

SpaceRox1244:
Add an additional inventory slot and adjust the rest of inventory to achieve a more dynamic system. The rig slot is an upper body parallel to the belt slot. Together, the belt slot and the rig slot form a task-oriented "front inventory" while changes to back bags form a spacious "back inventory".

This is just me bringing the doc overview forward into the PR description. But yeah I think this has good potential for role identity and creating an interesting relationship between the players different inventory spaces, with everything forming a complete picture rather than effectively tacking onto the back bag's central space.


slarticodefast:
Hey, I brought this up for a quick maintainer discussion and the general consensus was that our inventory can already be quite unintuitive, especially for newer players. Adding additional slots or conditions like doafters makes this more complicated and is not something that would fit into the upstream game design.
Thank you for your proposal.

SpaceRox1244:
Wait so you like, had the discussion on my behalf? Where's like, any of the actual engagement in the PR thread?

SpaceRox1244:
To be clear, I'm just saying that I feel left out of the dialogue here.

PR #228, opened 2024-05-31
Salvage Proposal

EmoGarbage404:
Does what it says on the tin

solidifies salvage's place in the game and clarifies a bunch of existing mechanics to work better in tandem

### DIFF
--- a/src/SUMMARY.md
+++ b/src/SUMMARY.md
@@ -200,7 +200,8 @@ Space Station 14
 	- [Cargo/Salvage](en/space-station-14/departments/cargo-salvage.md)
 		- [PR Guidelines](en/space-station-14/departments/cargo-salvage/guidelines.md)
 	
-		- [Proposals]()
+          - [Proposals]()
+            - [Salvage Proposal](en/space-station-14/departments/cargo-salvage/proposals/salvage-proposal.md)
 
 	- [Command](en/space-station-14/departments/command.md)
 		- [PR Guidelines](en/space-station-14/departments/command/guidelines.md)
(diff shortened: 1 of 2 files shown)


AJCM-git (approved):

RiceMar1244:
Everything here sounds really cool, but can I suggest that blueprints might have more flexibility and novelty as usable sheets of plans than lathe upgrades? The concepts definitely are not mutually exclusive, but if they say, opened a crafting menu that had unique recipes when you activate them, I feel like they could fit in a lot more places and enable more crafting.

They could start both item and structure recipes, use a greater variety of materials, be made available from more sources in the game (like maintenance scribbles in a locker and nefarious syndicate schematics in the uplink).

Ultimately they are definitely not mutually exclusive. They'd be totally independent functionalities and if you wanted the schematic to not offer one or the other you could just say some shit in the description like "these instructions do not seem machine legible/human legible". I only mention any of this in case one implementation would restrict the other.

AjexRose:
somewhat related to what RiceMar said, but blueprints were once being worked on but stopped after

RiceMar1244:
Lool yeah that was an entirely different and admittedly bad implementation thought that thing is a nightmare PR

EmoGarbage404:
i don't think the specific mechanics of blueprints is particularly important to this doc besides just the basic concept of "you get a replenishable non-sci way of creating things"

it could be expanded or changed however down the line and i don't think it matters much.

RiceMar1244:
Yeah you right you right. Just didnt wanna step on any toes, cheers

PR #244, opened 2024-06-23
Add information about Rider EAP

ferluci:


PJB3005 (changes requested):
EAP isn't always free, you should update this to reflect that.

Errant-4:
Thank you for your contribution, but it appears that at some point the target doc has already been updated with the more up to date info of Rider being free for noncommercial use. As such, I'm closing this PR

PR #245, opened 2024-06-23
Security improvements

Killerqu00:
A proposal with several security issues being put in a single design doc.
This doc proposes:
- Rework to tazers, adding them back to security;
- Improvements for alert code system;
- Gameplay implications for having high alert code set;
- Some other minor tweak suggestions (that influence other departments at times, that's why I think they should be in a design doc too).

### DIFF
--- a/src/en/space-station-14/departments/security/proposals/security-improvements.md
+++ b/src/en/space-station-14/departments/security/proposals/security-improvements.md
@@ -1,0 +1,120 @@
+# Security Improvements
+
+| Designers | Implemented | GitHub Links |
+|---|---|---|
+| Killerqu00 | :x: No | TBD |
+
+## Overview
+
+This design document reviews main problems with security department that are currently present in the game and proposes fixes to them. This document will not go over recent (at the moment of writing this document) changes in rules and Space Law due to them being changed and polished at the moment.
+
+## Background
+
+This design document is a result of many discussions about current problems on both Discord and forums. Some links:
+
+[Playing security is miserable](https://ptb.discord.com/channels/310555209753690112/1237264749281021952) on Discord
+
+[Prisoner and security feedback](https://ptb.discord.com/channels/310555209753690112/1231862578536251434) on Discord
+
+[Security in general is miserable](https://ptb.discord.com/channels/310555209753690112/1218720748097503252) on Discord
+
+[Security tweaks (Rework?) feedback](https://ptb.discord.com/channels/310555209753690112/1203731006331560027) on Discord
+
+[Security department is in extremely bad state](https://forum.spacestation14.com/t/security-department-is-in-extremely-bad-state/8147) on forums
+
+[Why is security so miserable to play?](https://forum.spacestation14.com/t/why-is-security-so-miserable-to-play-what-can-be-done-about-it/7708) on forums
+
+## Hall Interactions
+
+Hall interactions are actions that security performs to you usually during the blue alert or if there is something suspicious about you. It may be a random search, a question like "Why are you carrying a spear as a passenger?" or a peaceful arrest. Non-peaceful arrests and chases will be covered outside of this section.
+
+In general, there is a big problem with hall interactions: escalating the interaction to a fight is better when your opponent is typing something. This is very bad, because it means that it is better to magdump/stunbaton/run from them instead of actually going with the situation.
+
+There are 2 solutions I see here:
+
+1. Message interruption (the "-OOF!" thing) should work not only on being stunned, but on any hit, melee or gunshot. That automatically unfocuses the window, which allows for instant action.
+
+2. Harm mode indicators. Here's how it could look:
+
+![Harm mode indicator](https://i.imgur.com/nSY76PA.png)
+
+## Chases
+
+Let's say your hall interaction has already escalated to a chase.
+
+Currently, there is almost no way for security to cut the distance between the pursuer and the escaping person. Disablers exist, but they have low charges and are heavily ping-dependant. The solution to this problem is some sort of ranged slowdown for security. I suggest to implement this via reworking tazers and adding them back as a T1 research.
+
+Instead of insta-stunning, tazers will now apply 50% slowdown for 5 seconds on a target, unless they are wearing armor/hardsuit.
+This introduces better tools for security to engage in chases, while still leaving room for counter if you are an antagonist.
+
+As a result, I believe that this will shift the chases into a state where it's no longer always a better choice for a suspect to just run away, which will lead to more issues with security being solved (or at least attempted) to be solved verbally, or (for antagonists) using something else instead of magdumping the closest security officer upon being asked a question.
+
+## Isolating areas
+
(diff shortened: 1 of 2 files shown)


ninruB:
Security has plenty of ways to cut the distance between the pursued and them, flashes, flashbangs, bolas (which are effectively what you propose for the tazer rework), you also naturally shorten the distance because of how airlocks function and how the station is close-quarters, and more situational or chance-based methods.

I don't see a reason to reintroduce tazers back to sec, especially as they are presented in this doc.

Killerqu00:
> Security has plenty of ways to cut the distance between the pursued and them, flashes, flashbangs, bolas (which are effectively what you propose for the tazer rework), you also naturally shorten the distance because of how airlocks function and how the station is close-quarters, and more situational or chance-based methods.
> 
> I don't see a reason to reintroduce tazers back to sec, especially as they are presented in this doc.

i decided to cut out more deep analysis on the chase part, but here's some points on this:
- Airlocks work in favor of escapee, since escapee can hide behind airlocks with access;
- Flashes were made useless due to welding masks being pretty easy to obtain. However, this might change if masks get reduced FoV;
- Tazers in the form that was proposed in my doc were already PR'd and (IMHO) close to being approved, but author decided to close the PR.

ninruB:
> * Airlocks work in favor of escapee, since escapee can hide behind airlocks with access;

Heavily depends on the exact chase situation, and typically the secoff will have enough time to squeeze in through the airlock, the secoff might have access or the pursued might NOT have access, maps like marathon have security checkpoints leading into department areas everywhere.

Airlocks passively slow the pursued down because they do not open instantly and most players bump into them rather than in-world interact from the further-most point of range.

> * Flashes were made useless due to welding masks being pretty easy to obtain. However, this might change if masks get reduced FoV;

Flashes are absolutely not obsoleted because some 1-5 people on-station (out of 80) choose to wear weldermasks, they still are extremely strong to the degree of being borderline overpowered.

Now I'll note that disablers are still very viable even on high ping even just to slow the pursued down, if not outright stamcrit them. I, personally, as a high ping (200-300ms) sec main, do not see any issue with using them.

Stinger grenades exist, just thought I should point this out as a FTR.

Neither do I think that chases should be trivialized, it's good if the suspect/pursued gets away because the secoff failed to catch them despite their toolkit being specialized for this; right now it's already very very well-oriented towards chases.

Killerqu00:
> > * Airlocks work in favor of escapee, since escapee can hide behind airlocks with access;
> 
> Heavily depends on the exact chase situation, and typically the secoff will have enough time to squeeze in through the airlock, the secoff might have access or the pursued might NOT have access, maps like marathon have security checkpoints leading into department areas everywhere.
> 
> Airlocks passively slow the pursued down because they do not open instantly and most players bump into them rather than in-world interact from the further-most point of range.
> 
> > * Flashes were made useless due to welding masks being pretty easy to obtain. However, this might change if masks get reduced FoV;
> 
> Flashes are absolutely not obsoleted because some 1-5 people on-station (out of 80) choose to wear weldermasks, they still are extremely strong to the degree of being borderline overpowered.
> 
> Now I'll note that disablers are still very viable even on high ping even just to slow the pursued down, if not outright stamcrit them. I, personally, as a high ping (200-300ms) sec main, do not see any issue with using them.
> 
> Stinger grenades exist, just thought I should point this out as a FTR.
> 
> Neither do I think that chases should be trivialized, it's good if the suspect/pursued gets away because the secoff failed to catch them despite their toolkit being specialized for this; right now it's already very very well-oriented towards chases.

I can agree with flashes and partially with disablers, but airlocks are being used exactly like that.

Anyway, it's not the most important part. Other changes in the doc touch on airlock access issues and security checkpoints, too.

thebadman4662:
Some kind of bola launcher smaller than 4x4 pneumatic cannon for sure would be nice, throws do not inherit momentum so ironically bola best use is not for chase at all but ambush or debuffing/stunning enemy that holds their ground or pursues you.

ninruB:
I don't think this ultimately solves any "real" security issues. The forum/discord threads you've mentioned barely touch the design points you address in this doc.

* Message interruption as proposed is neat, but I'm pretty sure that's how it works right now.
* Harm mode indicators don't solve any issue as nothing stops Urist from turning on combat mode 0.3 seconds before hitting the secoff or vice-versa.
* Holographic projectors as proposed are, again, neat, but could use some downside for the user like having a slow upon trying to move through them.
* I don't think that handing out access based on the alert level is a good idea. This takes out even more interaction between heads, crew and sec.
Ideally you'd separate mass emergency access from alerts and give it to its own separate console in the bridge or something alike that.

Killerqu00:
>     * Message interruption as proposed is neat, but I'm pretty sure that's how it works right now.

I thought so too, but apparently not, or at least it is inconsistent between different instances of damage, because I tested that with throwing knives.

>     * I don't think that handing out access based on the alert level is a good idea. This takes out even more interaction between heads, crew and sec.
>       Ideally you'd separate mass emergency access from alerts and give it to its own separate console in the bridge or something alike that.

I don't think that security constantly banging on the door (for example) of cargo while there is a dragon in there is "taking out interactions". They _could_ call a captain, HoP or QM, but they might be dead. In general, I think that this is just a redundant thing and should be migitated by having sec get soft AA on red. It kinda makes sense for a person that has a job to protect a department to access it at least in case of emergency.

src/en/space-station-14/departments/security/proposals/security-improvements.md
```diff
@@ -0,0 +1,120 @@
...
+
+For the sake of simplicity, I propose that on LRP, any emergency allows for searches or using high-grade security gear. On MRP, however, security would have to follow this table: if there are no emergencies allowing for random searches, security cannot do them.
+
+This change is actually pretty minor: command already announces what is going on when changing alerts currently. This change will allow for more context for latejoiners and for crew that didn't get the context of said alert level raise for some reason.
+
+## Alarm Consequences
+
+For the sake of modularity, this will be discussed separately from the Alert Codes rework, using the current alert system. 
+
+Currently, there are zero consequences for abusing alarm system. You can just set the alarm to blue upon any minor crime being committed in order to give security a random search pass.
```
deltanedas, line 100: wont this just make the problem worse

1. hos stubs toe on the door entering the bridge
2. code blue for the next hour
3. sec gets free guns as well as free search pass

deltanedas (changes requested):

src/en/space-station-14/departments/security/proposals/security-improvements.md
Killerqu00: That's valid, I should probably add something else to blue alert

Cojoke-dot:
meh, harm mode indicators is kinda... ehhh... people don't use harm mode for harming people most of the time(sspppiiinnnn), and if someone is about to attack you they are just going to toggle it on when they start, not when it would be meaningful.

Taser change... is... is that not the point of the disabler? Feels like there is no point in slowing someone if I also have a tool that slows and stuns them that also works on hardsuits.

The alert level changes feel like they don't make much sense gameplay-wise. There is no incentive to be on green, it's an active debuff to the station. Yellow is nice, I like that one. Red just feels like such a reason to rush red alert then never turn it off, "I got slipped, go red so I can have AA"

Killerqu00:
draft to make some edits and additions

PR #246, opened 2024-06-29
Voting Cartridges

IProduceWidgets:
Proposal for three new cartridges all focused on voting via your PDA.


IProduceWidgets:
There's implications for revs here, that personally I think are both funny as hell, and solve the hiding last command member issue. I did not however dive into them too deeply.

deltanedas:
revs could turn into among, impostors voting out the last crewmate :trollface:

AJCM-git (changes requested):
By design this should be something available to every crew member on spawn, but there's a limit on how many programs we can have by default before using PDAs becomes clunky, this is why default pda programs like these should have a very distinct and useful use-case, but this isn't the case here since the purpose of these cartridges are pretty niche can be fullfilled by other programs.
What i think would be a better way to do this is making it so you can post polls in news, then make the captain announcements post news and figure out a way for revs to do the same. I feel like this would be a much cleaner way to do the same thing while allowing other possibilities.

IProduceWidgets:
Hmm.
My only issue with integrating the Democracy cartridge into the news is the pagenation in the news app is not very good for this use case in my opinion. Changing it to another type of scrolling I don't think would fit the news releases however.
(Also I think news app currently has some gnarly open issues requesting it mostly be reworked.)

I wouldn't mind the Democracy cartridge simply being deletable since it can be reinstalled by talking to the HoP though. That could be a reasonable compromise if it seems to take too much realestate on the PDA or proves impractical to integrate with the new app.

IProduceWidgets:
Here's a quick UI mockup I think makes a lot more sense than the news reader pagenation.
image

AJCM-git:
Sorry for the late response, missed the github notif

> Hmm. My only issue with integrating the Democracy cartridge into the news is the pagenation in the news app is not very good for this use case in my opinion. Changing it to another type of scrolling I don't think would fit the news releases however. 

Pagenation is fine, the problem is that there is no central list of all the news with a search bar in the UI, that shouldnt be that hard to add.

>I don't know how to feel about mixing the reporter/librarian publications with stuff intended to be from command. I feel it kinda cheapens the idea that this adds some agency to command roles, though I might not feel that way if announcements or broadcasts also displayed in news app.

Yeah civilian polls would be non-binding and only serve to enrich player to player interactions, only the ones from commands and revs would matter. Tho it may get confusing distinguishing between command and non-command polls/news if we keep the current UI but thats easily fixable as well (we can use coloring, making the author section more detailed, etc) 
 
> I wouldn't mind the Democracy cartridge simply being removable since it can be reinstalled by talking to the HoP though. That could be a reasonable compromise if it seems to take too much realestate on the PDA or proves impractical to integrate with the news app. Honestly, you should probably also be able to remove the news app if you wanted to, if only to add more things for HoP to be papers-pleased with.

This is the kind of feature that should be allowed by default and I dont think allowing the removal of default apps is a good idea since it gives room to way too much user error that just unintentionally worsens their experience (someone not that familiar with the PDA messing around and now not being able to access news or messaging and having no idea how to get them back). Tho i kinda dig that HoP idea of managing PDAs more directly, maybe having printed PDAs have no programs and giving a couple boxes of "default" cartridges to the HoP would be fine (if you truly want a custom PDA with certain programs you go to em)
 
> However, I don't even think we're terribly close to space on the home page of the PDA being an issue despite the two apps this would add to command pdas. The only other one I'm aware of in the works is PDA messaging, which could just as easily take the place of the notes app if we were to follow this line of thinking.

My issue isnt space, its usability, theres an usability cost for every program we add by default to the pda and every slot has a small value, if the use-case warrants it that small cost can be "paid", and thats the case for all default programs, in this case the feature can be *logically connected* to another program and thus i think it should be converged with that other program (they would need to share a lot of features anyways since if you want to make a poll with a lot of background information you basically need to make a news article, specially as we add more features to them like image support, more markdown, etc). 
The notekeeper merge with the messaging doesnt make much sense since you cant logically connect the two features (some people use personal chats to dump random things but thats more of an useful hack/trick and not their intended purpose, while its common to see polls in news articles). A better example and one i would agree with is merging messaging, news and polls in a single "Forum" program and that would be good but extremely hard to get right so its fine to keep some of em separate for now.

slarticodefast:
The author is an admin and was part of the internal discussion on this doc, but for transparancy I'll make a summary of that regardless:

- The maintainers like the proposal and it has conceptual approval, but the feature is considered minor enough that it does not need a design doc, so anyone is free to make a PR for it. Note that cartridges are currently frozen until Milon finished his work on making the pda UI predicted, see https://github.com/space-wizards/space-station-14/issues/8524.
- It was suggested to merge the three cartridges into one.
- Making votes could maybe be moved to a console mapped in the bridge instead.
- Some details of the demotion vote might needs some adjustments, but that can be further discussed once a PR is made. For better atomization it is probably easiest to first add the voting cartridges and then add the demotion feature in a separate PR once that is merged.

Since this feature does not require a design doc, I'll be closing this PR. Thank you for your contribution!

PR #261, opened 2024-07-19
Grab & Martial Arts

FaDeOkno:
Not really big design docs because they were requested by metalgearsloth.
I'm not sure if this enough, but I just can't find anything else to add.


ElPidor:
Interesting concept and realizzation
 ss14 does have a boxer role, for future pr's about different martial arts, this role is a huge playground

DocNITE:
Months ago...

PR #263, opened 2024-07-19
Changeling antagonist

Blackern5000:
Everyone knows the changeling, the space monster that eats people and then turns into them.
Based mainly on goonstation's changelings with a few elements from tg, just the armblade and stinger chemicals.

The main thing about changeling is that it round-removes it's victims rather than temporarily subduing them. Everyone is a victim, not just a select few people. To make up for this, the changeling has the ability to create ghost roles that assist it in it's goals. Creating ghost roles is optional for the ling, it is not obligated to bring it's victims back into the round.
The distinguishing feature about this version of the changeling is that it's humanoid form is it's unnatural form, it actually looks like a horrific flesh monster (depicted below), and it can choose to reveal it's true form as a combat mode of sorts to grant itself increased survivability and better offensive abilities.

pixil-gif-drawing - 2024-08-07T055243 323

Feedback is greatly appreciated.

### DIFF
--- a/src/SUMMARY.md
+++ b/src/SUMMARY.md
@@ -184,6 +184,7 @@ Space Station 14
 		- [Pizza Delivery Critter](en/space-station-14/round-flow/proposals/pizza-delivery-critter.md)
 		- [Rogue Drones](en/space-station-14/round-flow/proposals/rogue-drones.md)
 		- [Turf War](en/space-station-14/round-flow/proposals/turf-war.md)
+   		- [Changeling](en/space-station-14/round-flow/proposals/changeling.md)
 		
 - [User Interface](en/space-station-14/user-interface.md)
 	- [PR Guidelines](en/space-station-14/user-interface/guidelines.md)
--- a/src/en/space-station-14/round-flow/proposals/changeling.md
+++ b/src/en/space-station-14/round-flow/proposals/changeling.md
@@ -1,0 +1,186 @@
+# Changeling antagonist
+
+| Designers            | Implemented | GitHub Links |
+|----------------------|-------------|--------------|
+| alzore_ (Discord)    | :x: No      | TBD          |
+
+## Overview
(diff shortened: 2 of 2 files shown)


Cojoke-dot:
Aw, no changing hivemind... also kinda fucks your stealth up bad when your leaving husks of the people you eat everywhere and massive bloody puddles.

Cojoke-dot:
What about the goon hivemind, that was the best part about their lings in my option. Getting round removed by a ling did not bring you out of the game, it turned you into what is essentially a pai in a private coms line. Whenever the ling wanted they could split off their body parts and give the role directly to one of their hivemind members or just give their own body to one of them(so they could disguise as the person without acting different).

whateverusername0:
look good overall
also how many lings can be there at once? there's a maximum of 8 traitors right now, what about lings?

Djungelskog2:
im wondering, would they be their own gamemode or would they be integrated into traitor and such?

PotatOverlord:
If a headcrab went and took over a dead body, losing all the biomass and abilities, but then ate its old corpse should it regain all it's previous abilities and biomass?

mirrorcult:
if this is to be based on goon ling, their version of the hivemind/limb creatures is mandatory imo

mirrorcult:
by far the most interesting part of their design

Djungelskog2:
> im wondering, would they be their own gamemode or would they be integrated into traitor and such? considering we dont have the same dynamic gamemode 13 does

Asking again

Djungelskog2:
when we get more antags im guessing it will swap over to a more dynamic gamemode though? (its kinda unrelated but I want to know if its your intention with it)

whateverusername0:
@Blackern5000 as i can see you already have some sort of sprites ready for his frenzy form
so if you want we can work together to make it real [on this PR](https://github.com/space-wizards/space-station-14/pull/30094) 😃
i think the biomass and frenzy mechanics are cool and if we combine this design doc with whatever exists now we'll get perfect station 14 changelings

EmoGarbage404:
soldier

whateverusername0:
LET'S GO

Reese1243:
YOOOO ITS FINALLY HERE

BombasterDS:
I think the most important thing that needs to be done - action bar that will remember position of abilities that you make. On goob you just don't want to morph to another person because it'll mix all your actions.

PR #266, opened 2024-07-20
New race resomi

pofitlo-Git:
Add new species - Resomi


VasilisThePikachu:
Hello and thank you for the proposal. Unfortunately after some discussion with the maintainer team, we feel like this proposal does not meet our standards in it's current implementation for the wizard den (upstream) branch of servers.

You are more then welcome to write up a new proposal, assuming the following design problems below are addressed in some way. If you got any more questions, you may contact a member of the [species/character work group](https://docs.spacestation14.com/en/general-development/work-groups.html)

Here are some maintainer opinions:
1. The white flash goes against our goal for accessibility. Originally this was changed since the white flash was quite blinding. Forcing it for a species sucks.
2. Their current traits just make them avali, Avali in general are an open species, we don't need to legally destict them especially since they aren't any meaningful changes to separate them.
3. This species is at a giant disadvantage in combat situations. Especially in the health department when they don't have any meaningful upsides to them.
4. They are though, very powerful at tiding with the night vision trait of theirs and ignoring tables. And the other nerfs don't really help balance that out.
5. The pickup idea should probably just be a global system instead of something only one species gets to expirience. It would also make security play against them as easy as just picking them up.

Avalon-Proto:
I will put this here, a server that may accept this species is Delta V, You migh tneed to make a few extra clothes/suit sprites, but our species are more varied so might be worth a shot

PR #268, opened 2024-07-21
Hitman antagonist

Killerqu00:
This proposal describes a new stealth antagonist focused on assasinations of targets (often in specific ways).


deltanedas:
the no murderbone thing would be virtually impossible to implement, theres so many different ways to directly and indirectly people you just cant track it. instead i think relying on lore would help, say you are a contractor NT hired to weed out bad apples and nobody else. NT wouldnt be happy with you killing off random non targets

Killerqu00:
> the no murderbone thing would be virtually impossible to implement, theres so many different ways to directly and indirectly people you just cant track it. instead i think relying on lore would help, say you are a contractor NT hired to weed out bad apples and nobody else. NT wouldnt be happy with you killing off random non targets

While it would be virtually impossible to implement to account for every kill method, it would be possible to track at least some methods. Mass sabotage is currently being enforced via rules for all antagonists, so I think the plasmaflood scenario is unlikely to happen anyway.
If that will be a problem, I could easily replace an objective with something that cannot be achieved with mass sabotage.

PR #276, opened 2024-07-30
Pet cat ghost role

SpiritFoxShikigami:


IProduceWidgets:
You edited the proposal template. You're supposed to copy it to a relevant directory and edit it in place there.

IProduceWidgets:
I think Id prefer this stay a product of using cognizine rather than an item you can just acquire.

SpiritFoxShikigami:
Sorry about the template edit, I will change it later on in the day when I can.

I agree that cognizine does this job in a way already of making a cat sentient, but the biggest issue I see very often is that making it requires collaboration between botany and salvage.

You have to precisely ask those two and hope that they cooperate abd do not forget about your requests. In addition, you also need to mix the chemicals wjen you are not already occupied making medicine as a chemist.

My point is, is that this is something someone has to do specifically to enable a ghost role that can only walk amd talk but do nothing more than that.

A non-cognizied pet cat gets more features to work with to enable a more fleshed out relationship with an owner and making the ghost role a ton more interesting than waiting for someone (if it even happens at all) to make cognizine and having an underestimating gameplay loop.

Please reconsider the thoughts on it. Pet cats do not need to exist in mass and it's like having 1 to 3 is enough on one round to make people happy having a more meaningful ghost role other than just making rounds and watching.

The collar idea can also tie in to do that to existing cats instead of using a teleporter item, which is also good in my book. Pet cats do not have the ability to talk with out cognizine, but their sentience is still important for mentioned reasons above.

SpiritFoxShikigami:
Sorry about the template edit, I will change it later on in the day when I can.

I agree that cognizine does this job in a way already of making a cat sentient, but the biggest issue I see very often is that making it requires collaboration between botany and salvage.

You have to precisely ask those two and hope that they cooperate abd do not forget about your requests. In addition, you also need to mix the chemicals wjen you are not already occupied making medicine as a chemist.

My point is, is that this is something someone has to do specifically to enable a ghost role that can only walk amd talk but do nothing more than that.

A non-cognizied pet cat gets more features to work with to enable a more fleshed out relationship with an owner and making the ghost role a ton more interesting than waiting for someone (if it even happens at all) to make cognizine and having an underestimating gameplay loop.

Please reconsider the thoughts on it. Pet cats do not need to exist in mass and it's like having 1 to 3 is enough on one round to make people happy having a more meaningful ghost role other than just making rounds and watching.

The collar idea can also tie in to do that to existing cats instead of using a teleporter item, which is also good in my book. Pet cats do not have the ability to talk with out cognizine, but their sentience is still important for mentioned reasons above.

PR #264, opened 2024-07-20
Traitor variety (take 3)

TokenStyle:
wololo
making this for https://github.com/space-wizards/space-station-14/pull/30198

PR #284, opened 2024-08-16
Botany rewrite proposal

drakewill-CRL:
I spent a lot of time recently digging through Botany code, taking notes on how it worked and what could be changed. The most important part of this is tearing "Rewrite all of Botany" into 5 or 6 smaller parts that can be done separately.

TLDR version: Plant Growth and Mutations become separate systems. Harvesting becomes a component or effect. Plants become their own entity instead of a variable on PlantHolder. Add hooks for later improvements for interactions.

### DIFF
--- a/src/SUMMARY.md
+++ b/src/SUMMARY.md
@@ -245,6 +245,7 @@ Space Station 14
 
 		- [Proposals]()
 			- [Plant Genetics](en/space-station-14/departments/service/proposals/plant-genetics.md)
+   			- [Botany Rework](en/space-station-14/departments/service/proposals/botany-rework.md)
    		- [Librarian Gameplay](en/space-station-14/departments/service/proposals/theshued-librarian-gameplay.md)
       - [Joker Roles](en/space-station-14/departments/service/proposals/joker_roles.md)
 
--- a/src/en/space-station-14/departments/service/proposals/botany-rework.md
+++ b/src/en/space-station-14/departments/service/proposals/botany-rework.md
@@ -6,13 +6,13 @@
 
 ## Overview
 
-This is a proposal to improve Botany (The system involving plants growing in-game) by making it support the ECS code pattern better than it currently does. This is a large task, but can be broken into 5 or so pieces to make this manageable. This is not a proposal to change Hydroponics (The department where people do the growing of plants), though most ideas for reworking Hydroponics require a more flexible Botany system.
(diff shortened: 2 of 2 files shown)


Partmedia:
Thanks for your thoughtful work and writeup on this rewrite proposal.

> This is not a proposal to change Hydroponics (The department where people do the growing of plants), though most ideas for reworking Hydroponics require a more flexible Botany system.

I think it would be better to consider what you might imagine hydroponics will become before you embark on such a large rewrite. Imagine that you rewrite all of botany's code without any user-facing changes. Then we decide that hydroponics needs to change in some user-facing way. Won't that likely require possibly significant changes in the botany code?

You might argue that "botany ECS" will make it easier to adapt to future changes, but it's not a guarantee that the abstractions you chose in this rewrite will be flexible enough to support it.

Given that, and that any rewrite typically causes:

1. A lot of breakage in downstream forks
2. A lot of implementation bugs that need to be ironed out

I do want to make sure that any rewrite is at least anticipating the needs of the future.

drakewill-CRL:
Here is [my Hydroponics write-up ](https://github.com/space-wizards/docs/pull/290). I did that second, because there will far more opinions on job content and changes than back-end code. In my hydroponics plan, the core changes to botany code would be putting a GrowthComponent on all plants that enables a second set of mutations to be applied, and being able to select and apply mutations or variables from grown produce to other plants. My hydro proposal could be done in the current system, though it would be easier to change and maintain after this proposal is in place. 

I did look for any previous proposals and discussions, but there were very few concrete plans or actionable ideas. I didn't intend to write a perfect system that would do everything anyone could ever want the first try, but one where changing it in the future didn't seem so daunting. Splitting out plant growth steps and mutations into their own logic is a big part of that, because then discussions on removing redundant stats and changing how plants grow can be done and implemented without necessarily changing the mutation code and vice versa.

TokenStyle:
Add your proposal to `src/SUMMARY.md` for example: [Example1](https://github.com/space-wizards/docs/pull/264/files#diff-a7e3a664e97aa06240065215cf96d786578e4609d0b42fdfe14be434bc97b22fR187) , [Example2](https://github.com/space-wizards/docs/pull/228/files#diff-a7e3a664e97aa06240065215cf96d786578e4609d0b42fdfe14be434bc97b22fR204)

Fildrance:
Since botany refactor was merged - it feels like largely stated goals are already achieved, so maybe we can close this?
@AJCM-git  close if agreed :godo:

AJCM-git:
There are still some ideas from the doc that could be applied but after the refactor it is heavily out of date with the current state of botany

PR #285, opened 2024-08-17
Added Soups & Stews proposal, a proposal to add even more to the kitchen

Moomoobeef:
### Finally making a doc for an idea I've had kicking around for probably close to a year.
tl;dr is that I am ~~blatantly ripping off~~ inspired by the ["evolved" recipes system in project zomboid](https://pzwiki.net/wiki/Cooking_(crafting)#Evolved_recipes), and I wanted to bring soups/stews to ss14 in a similar way!

### DIFF
--- a/src/en/general-development/feature-proposals/Soups_Stews.md
+++ b/src/en/general-development/feature-proposals/Soups_Stews.md
@@ -1,0 +1,19 @@
+# Soups & Stews
+
+| Designers | Implemented | GitHub Links |
+|---|---|---|
+| Moomoobeef | :x: | TBD |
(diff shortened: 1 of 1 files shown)


Moomoobeef:
crap I totally put it in the wrong directory. Hold on I'll fix that

Moomoobeef:
Closing this proposal. It's stale, and I don't currently have any intention of working on implementation soon. I still do want to do it Eventually™, but I need more C# practice before that is feasible. ~~That or I need to imprison a C# dev and lock them in my basement until they code it.~~

PR #287, opened 2024-08-20
adds more text about markdowns to Writing Guidebook Entries

UBlueberry:
Two reasons for this:
A: General knowledge so everyone knows what markdown tags there are and how to use them
B: To standardize markdown tag usage on the Wizden guidebook


juliangiebel:
Please don't call it markdown. The tags like this: `[color ...]` are called richtext or bbcode. The `#` is from guidebook code and is hacked in "markdown syntax"

UBlueberry:
whoopsie!

PR #288, opened 2024-08-20
fix errors in changeling.md

Blackern5000:
I just read through the doc again and realized I screwed a few things up. Fixes the following:
- "Rejuvenate" is now spelled correctly
- Hivemind is no longer a larger title than the rest of the ability sections
- Some hivemind abilities are now bolded properly


slarticodefast (approved):
Thanks for the cleanup!

PR #301, opened 2024-09-06
IC Languages

Hanzdegloker:
My proposal for IC languages which I have been wishy washy over for like a month tryna workshop.

Before anyone asks yes I know there is another proposal already up about languages, I've been thinking over my own version of this very same concept over for a long while now and seeing someone else beat me to the punch got me in gear to finally finish writing this. 

Happy to adjust as need be.


Everturning:
this is definitely under baked. the other doc has concepts for what languages would look like to someone who doesn't speak it. this doesn't. but that's besides the point. 

>  | Nar'sian | You know how people sound when they drink Cogchamp? Just that.

Why would they be speaking ratvarian? they're narsie cultists.

> | NT Crew | Galactic Common | Sounds like a combination of Mandarin Chinese and various forms of Spanish |
> |Humans | Galactic Uncommon | Sounds like a combination of Russian and English |

I don't like this because it might make people confused and ahelp people not speaking English in the English servers and the same for other languages. It should just be randomized characters for each letter. 

> RD, whom can understand but cannot speak Encoded Audio Language, just to make sure those silicons ain't up to no good...

I think in the future, the binary chip will be removed from RD's locker, and so this would really only benefit the RD if he was standing right next to a borg (assuming they still have the bug where they broadcast their messages over AI in whispers)

> Now, I would also like to state mobs such as mice and mothroaches, do not understand galactic common at all. Mothroaches are a little special and can understand moffic, but cannot speak it. Basically majority sentient and non-sentient creatures speak/understand their respective languge and thats it,

👎 👎 👎 . Let them understand common.

overall this seems to be just a worse form of the other doc imo. More tedium, less fun.

Hanzdegloker:
> I don't like this because it might make people confused and ahelp people not speaking English in the English servers and the same for other languages. It should just be randomized characters for each letter. 

I use sound like in the most vague of terms, I assure you it would not be hard to decipher that it's gibberish in game. Something I forgot to touch on in the document actually is that every language having a different font would be great for telling what language someone is speaking at a glance even if you understand it. Full Russian sentences in Galactic common on an EN server should be easy enough to spot.

> I think in the future, the binary chip will be removed from RD's locker, and so this would really only benefit the RD if he was standing right next to a borg (assuming they still have the bug where they broadcast their messages over AI in whispers)

I mean, if the binary chip is removed it still makes a neat situational trait. Borgs can still chat between each other in their borg tongue and two emagged borgs could plot something nasty in plain sight, never mind binary 
 
> 👎 👎 👎 . Let them understand common.

Not opposed to the idea, I already let the round start sentient pets do so. I just found the idea of playing as a mouse getting yelled at in a strange language amusing 


> overall this seems to be just a worse form of the other doc imo. More tedium, less fun.

Not sure where the tedium comes into play, as for fun I am not against tweaking the understanding common part.

dffdff2423:
I am against the concept of humanoid languages because they often lead to metafriending and because ss14 is at it's core a RP game and I think intentionally excluding people from the conversation is not really the best design for it. The main use case I can see for languages is insulting people without them being able to understand you- which is quite funny and I am all for. Languages can also provide a bit of character to species, but no more then what can already be done via custom emotes. I personally strongly think that the downsides of humanoid languages outweigh the upsides.

On a more technical side languages should probably depend on a full chat refactor due to the current state of chat code.

superjj18:
wizden would never allow race specific languages due to their stance on specism

DandiIion:
> I think intentionally excluding people from the conversation is not really the best design for ss14.

This is already thoroughly present in the form of radio channels/departments. Not a fair reason to veto imo. 

> often lead to metafriending

Having things in common is inevitable and I strongly disagree that it would lead to meta-friending, more like would lead to Incharacter-friending. Don't think its any different from salvage often becoming pals with their lil walkietalkies.
Same hat

Aisu9:
> Don't think its any different from salvage often becoming

or having a good interaction with a botanist/chemist/doctor/any job and start playing more of this job to be with them, you don't need a departmental communication channel or specie languages to form a group of friend, if you really want ot take on the language use helping forming metafriend groups then it's already be actively done by using English as our mean of communication in-game

LankLTE:
While languages do provide some gameplay opportunities, we feel that ultimately the only reason to use a species language is to exclude a group of players, of then those of another species or of another group. Due to our complete ban on speciesism and to prevent issues with metacliques / metafriending, we do not wish to have roundstart species have languages. The system would still be desired for wild animals and such, however. I'll be closing this document as it is entirely based on roundstart species languages, but if you wish to change it feel free to request it to be re-opened.

PR #302, opened 2024-09-10
Radial Meteor Shielding

august-sun:
Implementing meteor shielding for stations, primarily handled by engineering, but can be handled by the crew in a pinch.

### DIFF
--- a/src/en/space-station-14/departments/engineering/proposals/radial-meteor-shielding.md
+++ b/src/en/space-station-14/departments/engineering/proposals/radial-meteor-shielding.md
@@ -1,0 +1,48 @@
+# Radial Meteor Shielding
+
+Your title should convey the basic jist of your proposed changes. It should be short because the text will be linked in the sidebar.
+
+|AugustSun|✖️|TBD|
+
+## Overview
+
+Radial Meteor Shielding allows for the station's crew to protect parts of the station, using machines connected to the station's power grid. 
+These machines must be anchored, and will consume significant power upon receiving hits from meteors, and to a lesser extent space debris.
+
+## Background
+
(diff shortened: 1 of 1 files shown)


Everturning:
Will these be roundstart?

august-sun:
> Will these be roundstart?

If they are using containment fields, extra containment field pylons are usually mapped roundstart. Otherwise, it could be something to be researched by science, ordered, or both, possibly?

Everturning:
I'm mainly thinking about the meteor shield station objective from 13... 🤔

august-sun:
Oh yeah I don't think this is going to be that powerful, though if station objectives were added (which I am a HUGE fan of), that'd make sense!

thorjelly:
I had a different 'meteor shield' idea recently, which was also intended to fix the common evac bombing problem at the same time. Was surprised to see a meteor shield design already recently proposed, so I'll mention my ideas here. If they are different enough to warrant a new document, I can do that.

Name it a "Breach Shield". When anchored to a tile with power, it will increase the resistances of **all** structural tiles within a radius. It will be relatively short ranged. Crew must actively, proactively move and anchor the device to specific parts of the station in anticipation of specific meteor strikes. Total station coverage would be impractical.

Create a Meteorology Console at the bridge. This console shows a map with directional lines showing the path that upcoming meteors are going to take, and the time that they will strike. It will show, say, large meteors within 15 minutes of impact, medium within 10, small within 5. Bonus: Add the ability of crew with a HUD display and silicons to see AI pings. AI could help navigate crew to optimal positions to place the breach shield while AI uses the meteorology console.

Finally, add a Breach Shields to the evac shuttle as an extra measure against evac bombing. These extra measures of proactive structural damage control could be the solution that is needed to remove or reduce the vacuum damage nerfs, which as I understand it, were put in place "temporarily" to find a solution to how common evac bombing is.

Why? Adds a proactive rather than passive or reactive solution to meteors, and a solution to evac bombings. To me it is important that it gives the crew something to do instead of taking away something to do (spacing prevention, in exchange for being forced to always fix spacings). Gives an additional role to bridge and AI, meteor tracking.

PR #306, opened 2024-09-18
Add caddy and docker compose to robustcdn

VasilisThePikachu:
Caddy's reverse config may not need the [flush_interval](https://caddyserver.com/docs/caddyfile/directives/reverse_proxy#flush_interval) value, since it will automatically instantly flush if ``Content-Length`` is non-existent (my quick read thru the publishing code looks like it does not?)

~~Looking into why the dockercompose won't work without perm fixes~~

~~Encode works for the whole reverse proxy block and not just json~~


PJB3005:
Self merge smh my head

VasilisThePikachu:
> Self merge smh my head

smh my head you said it's fine

PR #307, opened 2024-09-21
AppearanceSystem Compliance

SlamBamActionman:
Ever since I decided to implement photography into SS14 I have ended up in a crusade against how sprites are currently implemented and modified in the game. This design doc is a culmination of multiple discussions and deliberations on how to generalize sprite modifications, how to achieve reliable sprite copies and a proposal on how to tidy things up across the codebase. Multiple systems would benefit from this standardization; this design doc aims to answer how to get there.


src/en/space-station-14/core-tech/proposals/appearancesystem-compliance.md
```diff
@@ -0,0 +1,108 @@
...
+
+The red arrows show "destructive" operations where you either can't know for certain whether an edit to a sprite has been made, or it requires iterating over all systems for all objects to be fully certain that you have copied visuals fully.
+
+### Proposed Solution, implementation
+
+To achieve the overview, the following implementation would be required:
+
+- Any game logic system that modifies sprites is split up into a game logic system and a visualizer system (i.e. inherits `VisualizerSystem`). The game logic system *sets* `AppearanceData` for the visualizer system to work with, instead of modifying the sprites directly.
+- All visualizer systems are connected to visualizer components (e.g. `ClothingSystem` have `ClothingVisualizerSystem` and `ClothingVisualizerComponent`). The visualizer system handles the visual logic, and the component on an entity indicates an entity should run the system.
+- `SpriteSystem`/`SpriteComponent` may not be accessed outside of visualizer systems to prevent game logic systems modifying sprites.
```
ElectroJr, line 79: If anything, this should just be a convention / rule of thumb, and not actually enforced via an analyser or whatever because there's probably always going to be some exceptions. Especially when it comes to systems/components that control animated sprite layers, where you don't want to dirty/network unnecessary data every tick (e.g., something like `RgbLightControllerSystem`)

src/en/space-station-14/core-tech/proposals/appearancesystem-compliance.md
```diff
@@ -0,0 +1,108 @@
...
+You can hardcode exceptions to this manually, but this is of course not tennable and is bad coding practice when a general solution should be used instead.
+
+### Proposed Solution, Overview
+
+In a perfect world, the process of setting and copying an entity's sprite would be the following:
+
+Setting a sprite: 
+1) An entity is spawned in and has a default prototype with sprite data set; this is effectively the "base" to start from.
+2) A game logic system (`ExampleSystem` & `ExampleComponent`) performs its functions and, as a result, requires the sprite to change.
+3) The game logic system sets a key-value pair in the entity's `AppearanceComponent.AppearanceData`, and ensures the entity has `ExampleVisualizerComponent`. This raises the `OnAppearanceDataChange` event.
```
ElectroJr, line 52: One of my gripes with this is that it will require networking redundant data, and I've generally been pushing people to using data in components instead of relying on AppearanceData unless its a server-side comp, or they want to use something like the `GenericVisualizer`.

I.e., if `ExampleComponent` is a shared/networked component, all the data required to update the sprite is already contained in that component, and copying the data to `AppearanceData` will just mean that anytime the component is modified it has to uneccesarily network that data twice. And currently its even worse as there are no delta states, so it always has to send all of `AppearanceData`.

src/en/space-station-14/core-tech/proposals/appearancesystem-compliance.md
```diff
@@ -0,0 +1,108 @@
...
+
+### Animations
+
+Animations (especially looping ones) would require syncing the start time/offset of the animation frame, so that if a new client loads the animating entity it can accurately set the sprite's animation. This needs to be replicated across multiple systems such as lights and movement as well. 
+
+### Examples of concrete uses
+
+- Remembering sprites at point in time
+  - End of round sprites
+  - Photography
```
ElectroJr, line 103: TBH I'm somewhat tempted to just move sprite component back to shared (with networking disabled by default) to make server-authoritative sprites & taking sprite snapshots much easier. Though that'd also have a performance impact on the server, though I have NFI if it'd be significant. It'd also be a giant refactor that'd require moving all visualizers to shared.

And while maybe somewhat overkill, sprite snapshots could also be obtained by actually serializing the whole entity & spawning a client-side dummy for rendering.

PR #312, opened 2024-09-30
Botany Overproduction Fix Designs

drakewill-CRL:
I wrote up ways make botanists get less produce out of plants without making it feel bad for players.


Luxzhv:
Looking at the Design Doc, I personally really like one-off mutations and space ants. 
One-offs Will give older Botanists a challenge of juggling seeds to try and get the mutation they want again after they saw it, reducing yield.
Space ants would be a fully new material and mechanic, allowing botanists to make sudo-itching powder.

Princess-Cheeseballs (changes requested):
This doesn't need to be a design doc. I think this issues presented are important to catalogue though. I'd redirect this file to Core-Tech Botany (you'll have to make a new file) and it can be merged.

PR #321, opened 2024-10-13
Support for non Space Station 14 Games

benev0:
This is a meta proposal which will need several other supporting documents. 

Please read the preamble and concept sections if you want additional info.


benev0:
I am opening this in hopes that this document may receive additional feedback.

benev0:
related: #152

PR #323, opened 2024-10-15
Paradox Clone

slarticodefast:
Draft until my cloning refactor is complete.
Suggestions for improving the proposal are welcome.


src/en/space-station-14/round-flow/proposals/paradox-clone.md
```diff
@@ -0,0 +1,37 @@
...
+
+A freak space-time anomaly has teleported you into another reality! Now you have to find your counterpart and kill and replace them.
+
+### Objectives:
+- Kill and replace  \<target\>, \<job\> to fix the time paradox. Remember, your mission is to blend in, do not kill anyone else unless you have to!
+- Escape to CentCom alive and unrestrained.
+
+## Possible optional additions
+- Give the paradox clone a chance to be friendly instead of an antagonist for more roleplay opportunities and to prevent clones from being seen as 'valid' to kill.
+- If the original is an antagonist, additionally give the clone the original's objectives as well. This makes them an even more perfect copy and may lead to some interesting interactions between traitors where you first cooperatere and then inevitably betray each other.
```
benev0, line 37: To avoid the valid on site issue for players to consider randomizing objectives and/or spawning without objectives, this will give paradox clones plausible deniability regarding their antag status. 

As for random objectives, these could could vary from less than lethal, survive, never be detained (cuffed), to the mentioned kill and replace. These would all depend on weather the station of origin for the paradox clone is assumed to be evil or if the clone has extra evil intentions.

Cojoke-dot:
> Feels like you could just rework terminators to become this with a few changes like making them look like the target, making them human rather than robot

Term was removed for a couple of reasons, one of those was newmed work(just a week away) so they didn't have to deal with its weird set of traits. Its also kinda lame to be able to figure out which person the clone is by dealing heat damage and burning of their skin.

GoodWheatley:
imo paradox clones should just spawn with everything the target currently has in their inventory instead of just the roundstart loadout. It not only gives the player more tools to accomplish their goal but also prevents metagaming by calling out the captain clone for not having a nuke disk.

BramvanZijp:
> imo paradox clones should just spawn with everything the target currently has in their inventory instead of just the roundstart loadout. It not only gives the player more tools to accomplish their goal but also prevents metagaming by calling out the captain clone for not having a nuke disk.

That will completely break game balance with one of a kind items like for example the disk.

GoodWheatley:
> > imo paradox clones should just spawn with everything the target currently has in their inventory instead of just the roundstart loadout. It not only gives the player more tools to accomplish their goal but also prevents metagaming by calling out the captain clone for not having a nuke disk.
> 
> That will completely break game balance with one of a kind items like for example the disk.

I don’t really see the issue with that though? It creates new opportunities for savvy traitors to accomplish their objectives and can make the fact that an important role got replaced more obvious.

slarticodefast:
> imo paradox clones should just spawn with everything the target currently has in their inventory instead of just the roundstart loadout. 

Besides the problems it might cause with items like the nuke disk I don't think this is technically feasible since afaik there is no way to completely copy an item with all its components, datafields and visuals. For clothing items I was able to simply use the original prototype they spawned from since they usually don't get modified after spawning.

Jezithyr:
> > > imo paradox clones should just spawn with everything the target currently has in their inventory instead of just the roundstart loadout. It not only gives the player more tools to accomplish their goal but also prevents metagaming by calling out the captain clone for not having a nuke disk.
> > 
> > 
> > That will completely break game balance with one of a kind items like for example the disk.
> 
> I don’t really see the issue with that though? It creates new opportunities for savvy traitors to accomplish their objectives and can make the fact that an important role got replaced more obvious.
Part of the issue with just fully copying the target's gear/inventory is that it creates a massive maintainability/bug risk. The best example of *why* designing a system like this is a bad idea, I'd like to point towards Sylas from league of legends. 

In league, each hero has 5 abilities (4 normal and 1 ultimate) many of which have mechanics that interact with each other. Sylas is a champion who can *copy* another champions ultimate ability. Sounds fine and cool right? Well, here's the problem. There are many champions that have mechanics that tie into their ultimate, which means that for *each and every one* of those champions, they need to create a *brand new* special-case version of their ultimate for Sylas because he doesn't have those mechanics. 

This means that instead of having to just maintain 5 abilities, they need to maintain 5 + *15* (or however many variants they need to make). Which ended up creating some *hellish* and hilarious bugs, but also had the side effect of forcing riot to *change how they design their abilities*. Because now every time they make a champion they need to think of "How will Sylas interact/use their ultimate" which just adds extra work and limits the design choices they can make.

TLDR: Something seemingly simple like "Just copy the targets items/components" turns into a massive maintainability nightmare because anytime you want to add a new item/mechanic you need to consider what would happen if it was copied. It sounds neat but the technical/maintenance cost (and inevitable hellbugs), are not worth what you get out of it.

Jezithyr:
I do really like this concept and it solves the biggest issues with exterminator. That said, I think it would be worth exploring if there was some way to get the same themeing of the exterminator because as of right now it's kind of just a weaker/less interesting changling 🤔. I think the reason why exterminators were so egregious when it came to murderboning is because they got access to their full power right away (and the power was a bit over-tuned) and they didn't really have a goal after completing their objective. Having some sort of "upgrading" mechanic would help constrain players from murderboning while still allowing them to "go loud" if they prepare carefully. 

I also think that having station ai now means that we can have a better objective for the exterminator... Killing your target would give you a code/upgrade that would allow you to instantly hack the station AI to convert it to your own side along with giving you *and the ai* a new objective... Exterimate or Convert all organics on the station into borgs. (If there is no station ai, a ghostrole for one will be created).

GoodWheatley:
> > > > making the reply chain readable
> > > making the reply chain readable
> > making the reply chain readable
> points about Sylas

I’d argue that the main difference between copying inventory and this example is that we already have support for this in the first place since the game’s items aren’t inherently unique. 

I can understand how Sylas’s ability can create a ton of edge cases that can break the game since every hero ability is supposed to be only usable by that hero, but I don’t see how simply copying the contents of what’s in the target’s backpack is going to crate the same problems since it’s just making a new set of those items. As far as I know the only item that could potentially cause issues for multiple of them existing of them on the station at once is the nuke disk, and that’s because pinpointers break when there’s multiple identical items to track which isn’t inherently a problem with the nuke disk.

I do agree this can create balance problems, but I think being immediately able to discern if a head of staff is a paradox clone or not by if they have their steal objective on them is ALSO a balance problem.

Jezithyr:
> I do agree this can create balance problems, but I think being immediately able to discern if a head of staff is a paradox clone or not by if they have their steal objective on them is ALSO a balance problem.

It's less of a balance problem and more of a meta-leakage problem and one that other antagonists/mechanics share. Unconditionally copying items is a bad idea in generally because it's an uncommunicated dependency, and would require that all future items be designed with the knowledge in mind that they *may* be copied. I'd argue that creating that restriction just for a very minimal edge-case is not worth the bugs/maintainability burden. 

> being immediately able to discern if a head of staff is a paradox clone
IMO there shouldn't be visible items/steal objectives that can't be stolen and are able to immediately distinguish between a "real" head and a clone. That's a larger issue with how we handle role differentiation and items.

If a clone gets caught by sec and questioned/searched, they're going to likely realize that there are *two* of the same people on the station. And with a little bit of gaslighting from the clone... It turns into a "who is the real one" situation, after all the clone could have stolen the gear of the real one...

GoodWheatley:
> IMO there shouldn't be visible items/steal objectives that can't be stolen and are able to immediately distinguish between a "real" head and a clone. That's a larger issue with how we handle role differentiation and items.

I was mostly referring to steal items that can’t easily be acquired by a paradox clone, like the captain’s laser gun. If you have two identical captains in security the real one can just flash their laser gun and instantly prove they’re the real captain. It is possible for the laser gun to have been stolen by a third party altogether, but if the antagonist design is relying on completely unrelated antagonists to do something to create social intrigue I think that’s a design failure.

> If a clone gets caught by sec and questioned/searched, they're going to likely realize that there are _two_ of the same people on the station. And with a little bit of gaslighting from the clone... It turns into a "who is the real one" situation, after all the clone could have stolen the gear of the real one...

I totally agree that a gaslighting war between the real person and the fake clone is a great scenario to happen, and ideally I’d like the antag design to be promoting this as much as possible, which is why I’d like to see it have almost no possibility for the clone to be discovered through metagaming what they do and don’t get. 

I do acknowledge that having multiple unique items could cause problems, but I think in this case it’s picking your poison: add maintainer burden to prevent metagaming or enable metagaming by not copying items. There could be an ideal solution that solves both problems, but I don’t have one at present.

Flareguy:
> I do really like this concept and it solves the biggest issues with exterminator. That said, I think it would be worth exploring if there was some way to get the same themeing of the exterminator because as of right now it's kind of just a weaker/less interesting changling 🤔. Which would be fine as a minor antag! But if you wanted to go bigger, I think we can do better 😏
> 
> I think the reason why exterminators were so egregious when it came to murderboning is because they got access to their full power right away (and the power was a bit over-tuned) and they didn't really have a goal after completing their objective. Having some sort of "upgrading" mechanic would help constrain players from murderboning while still allowing them to "go loud" if they prepare carefully.
> 
> With the introduction of StationAI I think we can do _even better_... What if _your target_ spawned with a special shutdown code for the StationAI, and a bonus objective "Prevent the synthetic rebellion". The goal of the exterminator would be ensure that the target dies and they get the code for themselves. **If the terminator kills the target** they will receive the override code, and be able to use it to turn the StationAI into a Malfunctioning AI.
> 
> Once this is done, the objective for both the AI, Terminator, _and any Borgs_ becomes _Starting the Synthetic Rebellion_. The goal would be to eliminate all sentient organics on the station _or convert them into borgs_. This would also disable the self destruct that the robotics console. Getting the "override code" _and connecting to the AI core_ would also unlock the "full power" of the exterminator along with adding the exterminator to the binary radio channel.

As interesting as this is, I honestly feel like a big strength of Paradox Clone is just how simple it is. You're spawned as a duplicate of another player and have to work out who's the real one - even just on the face of it, it's a great concept that leads to a ton of funny emergent scenarios.
I feel like the low-stakes nature is also just good since it means the crew doesn't have to sweat over the existence of a clone (though it is still probably something to worry about for security, especially if the target they're duplicating is important.) I think they fit more as a kind of strange, self-centered force rather then an antagonist that "sneaks past the defenses" to damage the station.

Djungelskog2:
> I do really like this concept and it solves the biggest issues with exterminator. That said, I think it would be worth exploring if there was some way to get the same themeing of the exterminator because as of right now it's kind of just a weaker/less interesting changling 🤔. Which would be fine as a minor antag! But if you wanted to go bigger, I think we can do better 😏
> 
> I think the reason why exterminators were so egregious when it came to murderboning is because they got access to their full power right away (and the power was a bit over-tuned) and they didn't really have a goal after completing their objective. Having some sort of "upgrading" mechanic would help constrain players from murderboning while still allowing them to "go loud" if they prepare carefully.
> 
> With the introduction of StationAI I think we can do _even better_... What if _your target_ spawned with a special shutdown code for the StationAI, and a bonus objective "Prevent the synthetic rebellion". The goal of the exterminator would be ensure that the target dies and they get the code for themselves. **If the terminator kills the target** they will receive the override code, and be able to use it to turn the StationAI into a Malfunctioning AI.
> 
> Once this is done, the objective for both the AI, Terminator, _and any Borgs_ becomes _Starting the Synthetic Rebellion_. The goal would be to eliminate all sentient organics on the station _or convert them into borgs_. This would also disable the self destruct that the robotics console. Getting the "override code" _and connecting to the AI core_ would also unlock the "full power" of the exterminator along with adding the exterminator to the binary radio channel.

This sounds kinda like chrono-legionare but good
Add a few more targets and add more tasks (hopefully ones that dont entirely single people out to make it less samey and round removal-y)

Kadeo64:
I like this. It's a unique minor antag with good RP opportunities and an incentive to not murderbone.. but I think that should be made abundantly clear. Especially because they are just crew essentially.

Whenever this is implemented, the paradox clone should appear on the end round screen clarifying that the clone was NOT the actual character, to make sure ooc salt doesn't get directed at the person who plays the non-cloned character normally.

Djungelskog2:
Oh yea, escalation and character role rules (or rules in general) should also probably be rewritten too, I can imagine lots of problem scenarios coming out of not just this but anything else added in the future, especially because they don't even work that well for what we have right *now*

slarticodefast:
> Something seemingly simple like "Just copy the targets items/components" turns into a massive maintainability nightmare 

Yeah, like I mentioned above this would be really complicated to code, impossible to maintain, and way out of scope for the PR. I already had to do the same thing for a few components (like traits or other character/antag specific ones) to get cloning working correctly. Those are in a simple whitelist I defined and only empty marker components in most cases. But copying *all* components, datafields, as well as visuals to another object is way too complicated.
For clothing items I currently spawn a copy of the item's original prototype, which works here since clothes are simple in function.

>With the introduction of StationAI I think we can do even better... What if your target spawned with a special shutdown code for the StationAI, and a bonus objective "Prevent the synthetic rebellion". The goal of the exterminator would be ensure that the target dies and they get the code for themselves. If the terminator kills the target they will receive the override code, and be able to use it to turn the StationAI into a Malfunctioning AI.
Once this is done, the objective for both the AI, Terminator, and any Borgs becomes Starting the Synthetic Rebellion. The goal would be to eliminate all sentient organics on the station or convert them into borgs. This would also disable the self destruct that the robotics console.
Getting the "override code" and connecting to the AI core would also unlock the "full power" of the exterminator along with adding the exterminator to the binary radio channel.

I think the paradox clone is perfectly fine as it is and we can directly port it from SS13. Its beauty lies in how simple it is and its focus on roleplay and social engineering. Not every single antagonist has to be a super powered, station ending force. And adding complexity just for its own sake is not automatically a good thing either.

> but I think that should be made abundantly clear. Especially because they are just crew essentially. IMO the best way to do this would to just apply normal admin rulings on paradox clone kills outside of their single target, so murderboning with no reason is just going to get you banned for being a shitter

I am planning to bring this up for a direct discussion with the admins when my cloning refactor is done and I undraft this doc. I would like to avoid relying on additional admin rule reinforment for this, as the "who and when am I allowed to kill?"  grey area would come with an additional workload for them. I hope that simply having the objectives as "kill your target" and "escape to centcom alive and unrestrained", as well as the lore (you are a normal crewmember in an unusal situation) is enough to discourage unneccesary kills.

> Whenever this is implemented, the paradox clone should appear on the end round screen clarifying that the clone was NOT the actual character, to make sure ooc salt doesn't get directed at the person who plays the non-cloned character normally.

Absolutely, this is a must-have and something I will include. Especially on MRP people know each other and I would not want to break up friendships because someone took a clone of you as a ghostrole and ended up being an asshole. Same with proper admin tooling so they don't accidentally ban the wrong person.

> Oh yea, escalation and character role rules (or rules in general) should also probably be rewritten first too, I can imagine lots of problem scenarios coming out of not just this but anything else added in the future, especially because they don't even work that well for what we have right now

Could you give an example of what you mean? If there is a problem, I can bring it up in the admin discussion.

Djungelskog2:
>Could you give an example of what you mean? If there is a problem, I can bring it up in the admin discussion.

Improved escalation (look to beestation for an example)
Swapping role status mid-game, like going from antagonist to nonantag
Free agents
Free agent escalation
Sec escalation
Murderboning/not doing objectives
More rule-role types or completely rewrite them because they currently all fall under categories that do not do anything to clarify them individually (like ratkings and skeletons or animals and crew)

Main issue I see here though is that the paradox clones still an antagonist after it kills it's original and it would be slowed to kill people outside of its specific target, unless there were entire pages of clarifications for one role, like other free agents and antags, so just rewrite them at that point to be less vague and have more finely tailored categories (once again, check beestation escalation rules)

Kadeo64:
> I am planning to bring this up for a direct discussion with the admins when my cloning refactor is done and I undraft this doc. I would like to avoid relying on additional admin rule reinforment for this, as the "who and when am I allowed to kill?" grey area would come with an additional workload for them. I hope that simply having the objectives as "kill your target" and "escape to centcom alive and unrestrained", as well as the lore (you are a normal crewmember in an unusal situation) is enough to discourage unneccesary kills.

i would hope it would discourage unneeded murderboning as well but some people will just see the antag/free agent part on a humanoid ghost role and get to stealing and murderboning like what has happened occasionally with unknown shuttle events

Everturning:
I want to mention the admin "Doppelganger" spawn that can easily be entered into the game to help with the technical side, because in my experience the doppelganger's goal is to kill the original.

Kadeo64:
> I want to mention the admin "Doppelganger" spawn that can easily be entered into the game to help with the technical side, because in my experience the doppelganger's goal is to kill the original.

sounds like /tg/'s paradox clone

Dead-Wasp:
I don't understand why clone must be an antagonist. We have evil twin antag on Corvax for a while before he was delete. That antag was really boring for playing against. People learned to find a clone very quickly. In addition, the target for murder was quite easy or impossible to accomplish. It could be a scientist who studies artifacts all round without leaving. Yes, it will be easy for you to kill him, but in order to replace him, you will have to study the artifacts yourself. Or it could be a bartender who is always in the bar with people. If an ordinary traitor can try to take him to the technical tunnels. But his clone will not be able to, because everyone knows that the clone wants to kill the original. Personally, even on that server, it seemed to me that it is better to make the clone not an antagonist. Like a skeleton. Make it so that the clone, having appeared on the map, does not have any tasks and is free to do what he wants as long as it does not spoil the gameplay for others. If you want, you can become the best friend and partner with your original and together escape from a nuclear explosion in the closet. If you want, you can, without being seen by the original, do dirty tricks, commit crimes so that the original, who does not understand what is happening and is not aware of the presence of a double, will have to answer for them.

VasilisThePikachu (approved):

src/en/space-station-14/round-flow/proposals/paradox-clone.md
```diff
@@ -0,0 +1,37 @@
...
+## Implementation details
+Cloning code will be refactored to properly copy traits, DNA, fingerprints and any other relevant details from the original.
+The clone will get a copy of the outfit and basic equipment (e.g. toolbelts or goggles) the original is currently wearing, restricted by a whitelist. If this causes any problems from a code perspective alternatively the original's roundstart loadout can be used. However, this would be less ideal since many players customize their character during the round.
+
+### Ghost role information:
+
+A freak space-time anomaly has teleported you into another reality! Now you have to find your counterpart and kill and replace them.
+
+### Objectives:
+- Kill and replace  \<target\>, \<job\> to fix the time paradox. Remember, your mission is to blend in, do not kill anyone else unless you have to!
```
Simyon264, line 32: I think there should be a mechanical reinforcement to the amount of people you can kill. Something like "temporal instability". Kill/harm too many people and you yourself take damage or get reduced max HP or a similar debuff. Theres plenty of people who just go all ham on the crew if they are free agents / solo antagonists.

Simyon264: Resolved. 
> regarding your comment, I'm not sure how good a "max allowed number of kills before you get gibbed" counter would work
You should be able to defend yourself when attacked
and in some cases you have to kill others, even as normal crew (nukies, npcs, someone attacks you for no reason etc)

Simyon264 (approved):

ScarKy0 (approved):

PR #262, opened 2024-07-19
Specter antagonist

EmoGarbage404:
A proposal for a reworked version of the revenant, dubbed the "Specter"

This antag is primarily focused on haunting and stalking players, building up their fear, until it can approach them and deliver a killing blow to claim souls. Souls are used to unlock higher tiers of abilities and the fear generated can be used to purchase them.

PR #325, opened 2024-10-19
Revenant V2

Warpzoned:
A proposal for a complete re-design of the Revenant, with more abilities and ways to make the crew's lives a living hell!

Drafting until I get back to this.


Princess-Cheeseballs (changes requested):
Hey if you're still working on this, we'd like some major changes made to it.
1. An overall focus less on features and more on actual design principles. It should be less about "What abilities should the Revenant have" and more about "What does a good Revenant ability look like"
2. We'd overall like it to take more inspiration from Emo's design doc (https://github.com/space-wizards/docs/pull/262) specifically in regards to having horror and sanity integrated into the design, as well as some of the general conclusions posted in that doc's closing integrated into this one.

Warpzoned:
@Princess-Cheeseballs srry for the wait but I've finally started writing some up-to-date (and up-to-par... hopefully) stuff in the doc as per your feedback👍

it IS, however, still incomplete, so if you could analyze what I've written thus far and comment on it so that I at least know whether I'm doing it incorrectly (if that's the case) that would be Awesome Asf (updated text ends at line 68)

/ / / I'll probably be finishing this up in the next couple days as well if life lets me, otherwise only next weekend

PR #337, opened 2024-10-31
Draft: Character selection UI redesign

SaphireLattice:
Aka properly describe this mess

<details>
  <summary>Somewhat heavy image of an UI sketch</summary>
  
image

</details>


SaphireLattice:
Guess who really really got used to autosave feature of VSCode and Obsidian? Yeah.

SaphireLattice:
This proposal focuses solely on the UI/UX side of things right now. Mechanics will definitely need to be changed to match, as well as some extra UI components made (probably, mostly drag and drop I imagine). DB is... whatever. That's a problem for later.

Anyway.

On mechanical side of things, this will require:
- Split antags that are not subversive but have their own dedicated spawn into own ROLES. With this proposal you would be able to make a dedicated character just for nukies. No more "weird, I think I know this person". Could have a "random character" loadout item for this and shit, honestly.
- Split silicons, aka cyborgs and AI, into their own "species" sorta thing. Because let's admit it, they are already. The fact that they are tied to the characters' jobs is an issue with current job and character system, and this will fix that as well. No clue how much code stuff would be required for this though!

Everturning:
I don't really understand what this document is doing. I understand it's still a draft and the image is just a rough sketch, but I still can't figure out what it's saying. I agree on some points, like that loadouts need to be highlighted more and antags should have priority/preference levels, but a lot of it is still murky.

PR #339, opened 2024-11-04
Add dev enviroment setup video

VasilisThePikachu:
image

https://youtu.be/EUGl_zNS6Uk?t=3

Want to ask if people are find with the embed or if maybe it should be at the bottom? Just a link?


VasilisThePikachu:
Got some approvals that an embed is fine. Lesta go

PR #340, opened 2024-11-04
Minor updates to adding-a-simple-bikehorn.md

maxpiepenbrink:
Just went through this as a first timer, addressed some minor things that seem maybe out of date here.

* Added some language about the `Components` subfolder which appears conventional now in the relevant `Sound` folder used in this demonstration.
* Noted that you must leave the `Component` suffix off when making your prototype yaml and referencing your new component. People would probably figure this out but I figure explicitness is king.


src/en/ss14-by-example/adding-a-simple-bikehorn.md
```diff
@@ -145,7 +145,7 @@ Those helpers can be thought of as handling **client click -> server** and **ser
...
 ```
 
-Under the `Content.Server` project, there's a folder called `Sound`. That seems like a good place to put our new component (and in fact, this is where the real EmitSoundOnUse is located). Let's call our version `PlaySoundOnUseComponent`. Note: if you just copy paste this code in, it may not work, as you'll need to import various classes. Your IDE can do this for you.
+Under the `Content.Server` project, there's a folder called `Sound`. This folder contains an apty named `Components` folder. That seems like a good place to put our new component (and in fact, this is where the real `EmitSoundOnTriggerComponent` is located). Let's call our version `PlaySoundOnUseComponent`. Note: if you just copy paste this code in, it may not work, as you'll need to import various classes. Your IDE can do this for you.
```
VasilisThePikachu, line 148: What is an apty? I don't think that's what you meant to write?

VasilisThePikachu (changes requested):

src/en/ss14-by-example/adding-a-simple-bikehorn.md
deltanedas: aptly

maxpiepenbrink: whoops, my bad

VasilisThePikachu (approved):

Issue #15, opened 2023-09-21
ECS doc is bad and also severely out of date


mirrorcult:
its super all over the place and disorganized, doesn't explain much about how ECS is actually used in the codebase, needs more links to other explainers, and general updating for the current state of the codebase

+ probably needs explainers on archetype ECS / relationships / struct comps since we are getting that very soon with arch!

Jezithyr:
Dupe of #82

benev0:
IMO:
This document has a lot of character which is distracting.
If I where to alter/re-write it this would be my list of targets
- ECS doc should be split up into multiple parts. ECS, ECS + Events, FAQ
  - If FAQ is known it should be integrated into the doc on a re-wright  
- Sort out the word soup (mostly the "todo" section oop section)
  anti OOP rhetoric obscures the main point: *Constrained Programing*
- out of order examples: events are introduced too early just so code can be shown on screen 
- remove first person and favor third or second person (exclusive)
  - remove exclamations
  - remove narration
- remove rhetorical questions

PR #343, opened 2024-11-11
RT ECS docs rewrite

benev0:
This rewrite of the ECS doc has the goal of simplifying explanations of the architecture pattern.

- split the document into atomic parts
- remove and replace composition over inheritance diatribe
- provide minor edits to current quality content (event patterns)

fix: #15


benev0:
The system example does not have a update system in it.

benev0:
Added update to code example, but it may be worth splitting the sample up so that it is present adjacent to each of the appropriate sections. seeking opinions.   

I would appreciate a read through by someone who is not me. If you find something that is difficult to read, verbose, incorrect, or flawed, then feel free to mark it with anything from "here" or "this" to a detailed explanation of an issue. The latter is preferred for incorrect technical information. 

If you do not understand ECS concepts before reading and do not understand during or after reading, your review is very valuable to me, so please post your review. The goal is to bring you up to speed on ECS; you are the target audience.

PR #349, opened 2024-11-15
Blob Design Doc Proposal

TiniestShark:
This is my attempt at a solid design proposal for the Blob when it makes the leap over to Space Station 14. I'm a huge RTS fan and in doing a lot of research it's hard not to find the original Blob's design a bit lacking in creating an interesting back and forth between the Crew and the Blob itself. 

The Blob in SS13 tends to snowball incredibly aggressively to the point it doesn't really create an engaging antagonistic situation if the crew isn't rushing it as hard as possible (Which while the chaos of the situation can be fun, it's pretty repetitive over the long term of a game compared to other antags.)

Ideally this version of the Blob will be more fun for people to play without over-complicating the base concept too much, or making it too alienating, while giving players more options for strategy rather than only the hyper-aggressive play style of the original.


TiniestShark:
> can the blob talk or will they be rendered defenseless to me calling them stinky

They can talk to the others in the Blobmind, but for the crew it'll probably just eat you.

Thatonestomf:
Nah it just sprays you with blob spore so that you are also stinky.....and turns you into a horrific blob zombie
Then it insults you using the blob overmind

walksanatora:
link to rendered: https://github.com/TiniestShark/docs/blob/master/src/en/space-station-14/round-flow/antagonists/Blob%20Proposal.md

walksanatora:
does the game support mutiple radio channels on the same key
eg: would it conflict having binary on blobmind on the same :b key

TiniestShark:
> does the game support mutiple radio channels on the same key eg: would it conflict having binary on blobmind on the same :b key

Ah right I forgot about that. I'll have to pick a different channel then.

Alpha-Two:
Given that Numpad 7, 8, and 9 are used for camera control, would the default controls for quick-moving between nodes be better placed in the hotbar or, failing that, Numpad 1, 2, and 3?

TiniestShark:
Hotbar would be useful as a side but hotkeys for camera controls in an overhead environment should always be the preferred, especially when you're going to be covering so much of the station at once.

TiniestShark:
kind of questioning if this was ever really worth the endeavor tbh.

TiniestShark:
I can't really articulate my issues with the original blob well enough to make a convincing argument (even with every video having the same grindy issues) and I wouldn't be able to code this anyway. it's whatever.

PR #350, opened 2024-11-15
Update docs with new label definitions

SlamBamActionman:
Updates the docs with the small label adjustments, as well as clarifies what `S: Approved` is intended for. Fairly small fix!


slarticodefast:
Fixes the labels in the guidelines to correspond to our newest changes. Looks good to me.

PR #352, opened 2024-11-18
Anomaly Core Extraction

SpaceRox1244:
Add a machine for regularly extracting anomaly cores from anomalies. These cores will now be used to craft a set of round-start available technology that is unique to each anomaly type. 

The intention of this concept is to grant individual scientists more agency over the direction of their individual research, to bring all characters surrounding the anomaly research location closer to the tech rewards system, and to move exciting tech rewards into their own rewards spaces to reduce their relevancy competition.

### DIFF
--- a/src/SUMMARY.md
+++ b/src/SUMMARY.md
@@ -236,7 +236,7 @@ Space Station 14
 		- [Proposals]()
 			- [XenoArch Redux (3MOArch)](en/space-station-14/departments/science/proposals/xenoarch-redux.md)
 			- [Xenobio](en/space-station-14/departments/science/proposals/xenobio.md)
-  			- [Anomaly Core Extractor](en/space-station-14/departments/science/proposals/anomalycores.md)
+  			- [Anomaly Core Extractor](en/space-station-14/departments/science/proposals/anomaly-core-extractor.md)
 
 	- [Security](en/space-station-14/departments/security.md)
 		- [PR Guidelines](en/space-station-14/departments/security/guidelines.md)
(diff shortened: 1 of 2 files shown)


SpaceRox1244:
> So cores without destroying anoms for use in proto-machines that are round start printable at the protofab. I like the idea but I want to understand the proto-machines more:
> 
>     1. Will there be different proto-machines that require different cores, take generic cores, or do different things when different cores are inserted
> 
>     2. Will extracting cores change the current anomaly traits? If I draw out a core will it affect its stability or randomize its trait or node?
> 
>     3. Is ACE going to be locked behind a sci-tech or be round start?
> 
>     4. What are some concepts for the proto-machines? Are they meant to be high-tier tech that takes cores, like a pair of speedboots that run of cores but deal damage on use, or are they meant to serve odder purposes that are not already done by other techs?

1. Any of these would be entirely possible and would be up to the developer.
2. That is available for discussion but I think anomaly pulses accelerating the formation is an interesting start for injecting danger and complications into this. Personally, I dont think Im too excited about the idea of its shuffling the trait as if it were a transformation particle.
3. Considering that it grants access to a whole new catalogue of tech production, it's not intended to be an unlocked thing. If it wasnt round start, itd likely be something people rushed, which isnt really engaging.
If it were round start and still researchable, it would probably be like any of the other research equipment for replacement purposes. Honestly though, id be quite happy with not having to unlock those either, but thats not the scope here.
4. These prototype technologies are not an even higher tier with even higher tradeoffs. I personally am not engaged by a trade-off based design philosophy. A pair of speed-boots that are even more speed-booty but need to do some other bad thing just sounds like decorated vertical progression. Some of the best examples for what I would make into prototype tech is actually some of our tier 3s. For example, you might use a bluespace core to craft a bag of holding.

SpaceRox1244:
Closing this for now because I do not think I would feel like working on it right now if it were an approved concept.

PR #355, opened 2024-11-29
Update docs for new privacy policy info API

PJB3005:
Part of https://github.com/space-wizards/SS14.Launcher/issues/194


VasilisThePikachu (approved):

VasilisThePikachu:
Although perhaps this should be mentioned that this is an option in the server hosting guide? Perhaps under the level 3 guide next to the tags one.

PR #360, opened 2024-12-11
Creating, Editing, and Deleting Station Records

Centronias:
# Design proposal for adding create, edit, and delete capabilities for station records.

[The design proposal itself is here](link todo after PR open)

Meta stuff:
* Since I included images, I made a subdirectory to hold everything in one place
* I put this under "command" because it's more or less a HoP specific set of changes


beck-thompson:
commenting here so I dont forget

beck-thompson (changes requested):
Hi! There was a big discussion on this but here is the overall summery:

1.) Right now its kind of weird that "crew" status is purely based on if you have a job icon or not. You touched on this in your document a little bit, but it would be good if borgs / AI had some kind of HUD icon that would indicate if a crew members ID matched someone who was on the station.

2.) There was a lot of worry that being able to edit / delete station records would be used to allow AI and borgs to valid hut suspected traitors (This also assumes point 1 is implemented)! This was an issue in ss13, so its probably best to not allow any sort of edit / deletion of station records.

3.) There also needs to be a new method to determine how much mail is spawned as right now it just uses the total number of people on the manifest.

Besides that it looks pretty good :partying_face:! I personally love the hops scanner that seems like a fun device :+1:

Pronana:
Hi, I remember Reisama and I had a conversation about issues with how borgs determine who is and isn't crew and an idea like this came up in our discussions so I'll give some feedback based on what we talked about.

I agree with Beck that you shouldn't be able to add or delete records for the stated validhunting reasons unless there are some heavy mechanical and in-game restrictions around them (To limit it only to redundant or false records, or a prisoner who is to be executed) even then it's probably better to make it so crew can't be deleted from the records. 

In terms of minimum requirements for an entry, I think having it be name only is fine for non-crew and visitors but if you want to add someone as crew you *have* to scan them with the biometric scanner. Mechanically it should just not give you to option to add someone to the crew if they have not been scanned only visitor or non-crew. It should also check to make sure there isn't a duplicate person in the system already to prevent duplicating entries on the manifest. 

This would make it so the only way you could add new entries is: 
> If entries were already deleted by event
> If it's a brand new person coming to the station i.e. Visitor, Centcomm, Nukie in disguise
> If someone DNA scrambles themselves

In addition, mail should only be sent to crew-members and determining how much mail is sent should take that into account (Maybe the manifest only shows people who are actually crew?). This should prevent HoP from stuffing the manifest to get more mail.

Lastly, a funny antag objective for ninja could be wiping or scrambling the manifest which makes everyone on the station non-crew as HoP oriented mirror of the "Make Everyone Wanted" objective.

Centronias:
Solid feedback, thank you.

The point about deleting records is definitely important. If creation is limited by "you have to be biometrically scanned" and "duplicate entries are somehow disallowed", then it makes sense that true deletion of records should not happen.
I'll do something for like... "alive and on station like normal"/"dead"/"missing"/"cryo'd" for flavor (and possibly for making it so mail isn't spawned for gone people?). This'd mean that there's no way to validhunt because there's no way to remove somebody's crew status once they have it.

To the point about crew-membership determination, I'd like some more guidance. Is the expectation that visitors / nuke ops _should not_ be crew until they're registered with the HoP? To me, this seems reasonable, and like a decent way for visitors to be incentivized to get registered, but I'm not very balance-minded, so I'd like some opinions.
I still think there's a lot of flavor available for having non-crew registered in the system (think a non-crew antagonist detained and getting registered so they can show up as detained on the criminal records), so perhaps it could be possible to add records of noncrew which don't show up on the manifest. Those non-crew records could perhaps be upgraded to crew records, but crew can never be downgraded to non-crew.

Further, is the intention with this that silicons ought to be able to just sense crew status, or would it still be linked to IDs being equipped? I only ask because there's already precedent for "NT tech can sense IDs being equipped" but not for "NT tech can sense somebody's DNA at a distance".

Pronana:
> To the point about crew-membership determination, I'd like some more guidance. Is the expectation that visitors / nuke ops _should not_ be crew until they're registered with the HoP? To me, this seems reasonable, and like a decent way for visitors to be incentivized to get registered, but I'm not very balance-minded, so I'd like some opinions. I still think there's a lot of flavor available for having non-crew registered in the system (think a non-crew antagonist detained and getting registered so they can show up as detained on the criminal records), so perhaps it could be possible to add records of noncrew which don't show up on the manifest. Those non-crew records could perhaps be upgraded to crew records, but crew can never be downgraded to non-crew.

This is a tough one since it intersects with space-law and silicon law. I agree that it's probably best to allow the option to add non-crew to the registry with lesser restrictions as a "just in case" and then just not count them in the manifest. Plus it also gives rewards for invader antags like Nukies to go stealthily and get properly registered with HoP since it will give them a definitive crew status. 

This is definitely something that would require testing and player feedback I think.

> Further, is the intention with this that silicons ought to be able to just sense crew status, or would it still be linked to IDs being equipped? I only ask because there's already precedent for "NT tech can sense IDs being equipped" but not for "NT tech can sense somebody's DNA at a distance".

This is something that I don't think Rei and I ever reached a conclusive agreement on other than "It should be manifest based not ID based." The specifics of if Borgs could see if someone was on the manifest, or if IDs had a way to display if they were valid is something that requires changes to silicon rules meaning it would require admin discussion. 

I personally had the idea of it being biometrics based where borgs identified crew through facial and voice recognition, with the ID as a fallback for if the crewmember was wearing a helmet or mask (a very common occurrence). This would make it so borgs are often not sure if someone is crew and should just assume people are crew until the AI tells you the 5 guys in bloodreds shooting up cargo are in fact not on the manifest. 

Either way, unless the maintainers or admins want it, I'd suggest leaving this as something open ended. While the changes you're suggesting to the manifest create exciting new opportunities for changing silicon rules in fun and new ways, it's ultimately a big enough change that it's probably best saved for another design doc. I think it's fine to give suggestions and examples as to how this design doc can lead to improvements to how borgs operate but it's probably better to keep your scope narrow and not get bogged down in details.

Centronias:
@beck-thompson + @Pronana , please check out my edits to the text to address the concerns you brought up.

PR #374, opened 2025-01-07
Security K9 unit ghost role

JustinWinningham:
First draft of Security K9 ghost role proposition


Cojoke-dot:
So how this works is thats it's a security Remi who can search people faster? The new job is just a secoff with a pet then correct? The K9 dog would be considered a familiar then, would the ghost role be focused on an item or the K9 person?

Sounds cool I guess, the sniffing could definitely be broken with how it's implemented. Would it detect contraband, chemicals, or explosives? Could it sense stuff in a storage implant? At the very minimum it should take about 10 seconds for the dog to sniff so people can react and it's not just constant contra checks on green.

JustinWinningham:
> So how this works is thats it's a security Remi who can search people faster? The new job is just a secoff with a pet then correct? The K9 dog would be considered a familiar then, would the ghost role be focused on an item or the K9 person?
> 
> Sounds cool I guess, the sniffing could definitely be broken with how it's implemented. Would it detect contraband, chemicals, or explosives? Could it smell stuff in a storage implant? At the very minimum it should take about 10 seconds for the dog to sniff so people can react and it's not just constant contra checks on green.

I asked people a few times and generally people thought it would be more appropriate to be a ghost role instead of a full job. But yes you would act like a department level familiar - or attached to a particular secoff. The ghost could spawn as an activation off a K9 handler one time use item similar to syndiecats or just as a regular random mob.

I hard agree that it could be broken, the sniffing would take time just like stripping to prevent abuse.

NoElkaTheGod:
You mentioned it being able to sniff food. I think it would be funny, if food took priority over everything else. That way you can carry a steak around, so that K9 would not be able to detect any explosives or whatever you have.

Also consider K9 with cogni.

src/en/space-station-14/departments/security/proposals/K9Unit.md
```diff
@@ -0,0 +1,78 @@
...
+
+1) Denoted using an new security hud icon that is only visible to K9 units to match existing styles. This modifier will only be visible on characters in a short range of the K9 to prevent cross-screen snipes on players who are out of scent range and as a balancing mechanic.
+2) The K9 unit must use a short 1 second 'sniff' action on a player and get a text message popup similar to that recieved when somebody is placing an entity into a players hands that tells the K9 if they picked up any foreign scent, including chemicals, reagents, or foods. "You smell explosives on this person!" / "you smell <chemical scent> on this person" / "you smell <food taste or scent>.
+3) Scent mechanic is too powerful all together and is not implemented.
+
+#### Bite
+
+Unlike most bites from other creatures, this would inflict a small (0.25 - 0.5) second slow on targets, this would combo with their speed to give them more versatility in aiding security staff, further aiding their support toolkit.
+Additionally, this bite would be stronger when wearing a k9 hardsuit (detailed in equipment section below) in exchange for lower speed and higher armor.
```
Errant-4, line 46: No stun, tackle or stamina damage. These have very significant combat applications, and historically they have been very disruptive.
However, we would be interested in something like this: a short-range pounce ability primarily designed for chases. The ability can be used to execute a ~2 tile pounce. If it connects with a mob, the dog "bites onto" the mob, essentially being dragged from then-on. This slows down the target as if they were dragging, but does not occupy their hand. The pounce should have a long cooldown or failing to hit a target should be punished with some sort of movement penalty such as slowdown/stun so it can't just be used to dash around all the time

Thus, this ability could be used to close the distance to a fleeing target if you are already close, and cause slowdown on them, allowing everyone else to catch up. In this state, the dog would be unable to bite, so the target has the option of starting to punch the dog until it lets go, but... is that wise, when this will probably cause the dog to let go and start biting back? Likewise, the dog has to choose how long to hold on and when to start attacking/disengage. Being dragged after the target would also, ironically, serve to shield them from shots coming from pursuers

(This ability would be a not insignificant thing to code, but probably worth investigating even without the pounce part)

src/en/space-station-14/departments/security/proposals/K9Unit.md
```diff
@@ -0,0 +1,78 @@
...
+This proposal will require some new sprite and sfx assets:
+
+Sprites:
+* SecK9 Hardsuit: icon + equipped overlay for SecK9
+* SecK9 Armored Vest: icon + equipped overlay for SecK9
+* SecK9 entity: Up/Left/Down and Dead
+
+Sfx:
+* SecK9 alert whistle
+* SecK9 bark (if desired to be unique from other dogs)
```
Errant-4, line 78: +Sniff sound, if used for tracking

Errant-4 (changes requested):
We had a bunch of discussion about this, here is my attempt to summarise the first round of feedback from various people, hopefully with only minimal bias from me

src/en/space-station-14/departments/security/proposals/K9Unit.md
```diff
@@ -0,0 +1,78 @@
...
+
+This proposal will require some new sprite and sfx assets:
+
+Sprites:
+* SecK9 Hardsuit: icon + equipped overlay for SecK9
+* SecK9 Armored Vest: icon + equipped overlay for SecK9
+* SecK9 entity: Up/Left/Down and Dead
+
+Sfx:
+* SecK9 alert whistle
```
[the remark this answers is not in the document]
JustinWinningham: was referred to as alert bark in other areas.

src/en/space-station-14/departments/security/proposals/K9Unit.md
```diff
@@ -0,0 +1,78 @@
...
+with only Det having a minor variation in loadout and some added utility for forensics. All three focus on patroling, detaining, and generally dealing with antagonists with Warden and HOS being the exceptions. A K9 unit would add great variety for sec players while also opening many roleplay opprotunities with a low MVP cost (as corgi/cerb and their behaviors already exist in code at least in part). The Main goal for the person playing the K9 is to give a security role where they are not expected to verbally interact with crew, which can be draining, irritating, demoralizing, etc. This gives security a way for players who need a break from the negative human interaction side of the department a way to continue playing there, while also giving new tools to combat the most destructive tools available to antags which currently have very few counters, and allow for a ghost to occasionally help security if they want, as current ghost roles are neutral to the station at the best in most cases. Further itreators or refinement could see K9 role given more abilities or more scents to detect and filter through (drugs? food? etc.)
+
+## K9 Officer Role
```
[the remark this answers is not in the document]
JustinWinningham: will update name in v2

src/en/space-station-14/departments/security/proposals/K9Unit.md
JustinWinningham: added for v2

src/en/space-station-14/departments/security/proposals/K9Unit.md
BramvanZijp: What if being effected by it also did some damage on successful connection, and more importantly, while under the effect of the ability, taking certain actions (Moving/running, etc.) causes the mob to take damage and/or increase bleed stacks? Given that if a dog has you in its teeth, attempts to pull away, swing the limb, kick back, etc., would probably result in further injuries due to causing severe movement of the very sharp teeth currently clamped into your skin. (Would also be cool if this reduced gun accuracy and/or increased the length of certain do-after bars, perhaps with certain stimulants or sedatives reducing/removing this effect, but thats probably overdoing it.)

Would also be a cool balance concept since theyll have to either do major damage to the pupper to force it to stand down (I feel like if they get stamcritted or regular critted the effect would end), cooperate and stop moving, or continue to flee and face the very painful consequences.

(The same concept why if somebody has a bladed weapon currently stuck in them, an EMTs best bet is typically to secure it in place. Removing it unplugs the hole resulting in blood gushing out, but having it move around causes massive internal injuries, I feel like the simular risk of injury would apply with resisting/moving against a dog bite.)

src/en/space-station-14/departments/security/proposals/K9Unit.md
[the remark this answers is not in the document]
Errant-4: But you have that in the next line

src/en/space-station-14/departments/security/proposals/SecDog.md
```diff
@@ -0,0 +1,72 @@
...
+* Lunge: Dash attack which force-drags the target for a short duration. Slows the dog significantly for the same duration, even if they miss.
+* Signal: Whistle-like callout / bark to notify handler of contraband.
+* Drag: Dogs can drag bodies and other objects like other humanoids.
+* No Slip Paws
+
+#### Cons
+
+* Unable to talk or use radio(dog speech)
+* Unable to detain or arrest (only bite)
+* Limited resource (Ghost role not garunteed, and hard to replace if killed)
```
slarticodefast, line 24:
```suggestion
* Limited resource (Ghost role not guaranteed, and hard to replace if killed)
```

Errant-4:
Thank you, we have voted unanimously to merge the proposal.
I'm personally looking forward to see how this tests on Vulture

PR #378, opened 2025-01-10
The Stationary Contraband Detector

Mixelz:
A very simple in concept machine that should be a breath of fresh air for contraband as a mechanic in addition to creating the potential for more roleplay opportunities.

Link to the initial [Discord Conversation](https://discord.com/channels/310555209753690112/310555209753690112/1326793952267079713) that spurred this PR into action!

### DIFF
--- a/src/en/space-station-14/departments/security/proposals/StationaryContrabandDetector.md
+++ b/src/en/space-station-14/departments/security/proposals/StationaryContrabandDetector.md
@@ -25,7 +25,7 @@ Additionally, it can be used as a method to more reliably indicate the presence 
 
 The intents of this implementation on when, where and why it should be used are as follows:
 1. As a starting point, it should trigger on any contraband outside of its owners department or always if it is syndicate contraband. An item in a dragged container cannot be attributed to a specific owner/holder and thus would always set it off without an owner being attached to it.
-2. It should be limited ONLY to high priority areas, there should not be a scanner on every hallway or maints shaft and its research restriction and difficulty to setup should be set accordingly.
+2. It should be limited ONLY to high priority areas, there should not be a scanner on every hallway or maints shaft and its research restriction & difficulty to setup should be set accordingly. For this reason it should probably only be located on EVAC shuttles themselves baseline so that they cannot be stolen and moved round start.
 3. It should not, by itself, be a reason to jail or detain someone. It should ONLY be used to raise a flag for a security officer to engage upon with roleplay, either via following up with a manual search or interrogating the suspect directly.
 4. It should be able to be worked around IF it is planned for ahead of time. It should not be easily "cheeseable" through methods such as Q-Dropping the item past the detector, throwing the item past it or using a simple implant to ignore it. These would devalue more involved methods of a traitor getting past it and the point of the machine itself.
 
--- a/src/en/space-station-14/departments/security/proposals/StationaryContrabandDetector.md
+++ b/src/en/space-station-14/departments/security/proposals/StationaryContrabandDetector.md
@@ -1,0 +1,62 @@
+# Stationary Contraband Detector
+
+| Designers | Implemented | GitHub Links |
+|---|---|---|
+| [Mixelz] Killerqu00 | ❌ No | TBD |
+
+## Overview
+The Stationary Contraband Detector is a one tile machine that will scan the inventory of any containers or people that pass over it. 
+If the detector detects any contraband, it will alert those in its vicinity via a loud sound and lights like that of any alarm.
+It will also alert security on their communication channel that the detector was triggered and who set it off, but *not* what item set it off; this effect is proposed to be achieved by having the machine automatically update the Records system whenever it is triggered.
+
+It should additionally have a small percentage chance of displaying a false positive result so that it cannot be a guaranteed method of confirming the presence contraband.
+
+## Background
+
+In the current state of the game, an end of round evac bombing is all but guaranteed and it its almost entirely impossible to stop unless you happen to have all potential agents already detained or very strong suspicion that a specific person will attempt to carry out the bombing.
+
+There are also very few methods of security being able to accurately confirm that someone has contraband; they must either visibly see them carry the item in their hands or must have prior reasonable suspicion that the target has it due to a prior encounter that would only be possible through contraband.
+
+A contraband detector, something that is a staple of real world security operations, would be a method to help alleviate these problems, add another layer of protection that traitors will need to bypass and would be a mechanical method of being able to detect if there is contraband in play while allowing sec to search a volume of people larger than themselves without grinding an area to a halt.
+
+Additionally, it can be used as a method to more reliably indicate the presence of a Storage Implant (I.E. Someone consistently setting one off despite having an empty inventory.), which has been a point of contention as Meta Knowledge due to the limited avenues of knowing when one is in play.
+
+## Design Intentions
+
+The intents of this implementation on when, where and why it should be used are as follows:
+1. As a starting point, it should trigger on any contraband outside of its owners department or always if it is syndicate contraband. An item in a dragged container cannot be attributed to a specific owner/holder and thus would always set it off without an owner being attached to it.
+2. It should be limited ONLY to high priority areas, there should not be a scanner on every hallway or maints shaft and its research restriction and difficulty to setup should be set accordingly.
+3. It should not, by itself, be a reason to jail or detain someone. It should ONLY be used to raise a flag for a security officer to engage upon with roleplay, either via following up with a manual search or interrogating the suspect directly.
+4. It should be able to be worked around IF it is planned for ahead of time. It should not be easily "cheeseable" through methods such as Q-Dropping the item past the detector, throwing the item past it or using a simple implant to ignore it. These would devalue more involved methods of a traitor getting past it and the point of the machine itself.
+
+## Addressing Mentioned Concerns
+> This would be a hard nerf to Storage Implants.
+
+I think it would make storage implants easier to detect, but it is an opportunity for more adaptive gameplay to emerge instead of using the implant as risk free storage.
+A very simple example could be holding major contraband in the implant while carrying minor contraband in a backpack to act as the scape goat.
+A competent Security could then ask for a quick recheck, sending the traitor into a panic for how to resolve this situation with their contraband still in hand.
+Effectively, we could roleplay as TSA agents and that itself sounds like it would lead to more fun situations than what we currently have available.
+> This would remove evac bombs from being possible.
+
+It would not since bypassing the detector is planned as part of its core intent.
+It WOULD make it much more difficult to achieve for the DaGD objective type, but it is already a trivial objective to accomplish and would be a massive QoL improvement for the Evac portion of a round.
+> Shitsec could abuse this and violate space law with a dubiously "legal" method.
+
+It would be no different if those same officers decided to randomly search people without the machine and would still be breaking Space Law regardless. 
+It would fall under being a new reason of probable cause and would be a massive boon for the players that are able to behave themselves.
+Its planned implementation is already quite restrictive, and while the potential of abuse is there the same can be said for many other mechanics of the game and I do not think this falls outside those already established boundaries.
+>  This would potentially nerf traitors too much and/or buff security too much.
+
+Considering its intended restrictions, I do not see many scenarios beyond Evac Bombing that this would be a major change for.
+It would be specifically alerting the behavior in regards to casually carrying contraband, which I feel is something that has not yet been dealt with in any direct capacity.
+It would also encourage more RP scenarios as mentioned above, which may not inherently be a buff/nerf to either faction depending on how the "Meta" evolves from it.
+
+## Potential Follow Ups/ Other Notable PRS
+Some potential follow up PRs to this device have been discussed, which will be noted below:
+
+### Locking Uplink when Evac arrives. 
+This would be a very appreciated PR to ensure its intent of evac protection is upheld, without this it would be trivial to bypass it for Evac bombing purposes.
+### Allow the device to be EMAGGED so it always triggers a false positive.
+This will not be part of the initial implementation unless a maintainer would approve of it due to the EMAG restriction/freeze.
+### Create a handheld variant for individuals to hold in their hands.
+This would allow it to be more flexible in its use, though it should probably come with even harsher restrictions such as being unlocked with late tier research or being part of the armory.
--- a/src/en/space-station-14/departments/security/proposals/StationaryContrabandDetector.md
+++ b/src/en/space-station-14/departments/security/proposals/StationaryContrabandDetector.md
@@ -9,7 +9,7 @@ The Stationary Contraband Detector is a one tile machine that will scan the inve
 If the detector detects any contraband, it will alert those in its vicinity via a loud sound and lights like that of any alarm.
 It will also alert security on their communication channel that the detector was triggered and who set it off, but *not* what item set it off; this effect is proposed to be achieved by having the machine automatically update the Records system whenever it is triggered.
 
-It should additionally have a small percentage chance of displaying a false positive result so that it cannot be a guaranteed method of confirming the presence contraband.
+It should additionally have a small percentage chance of displaying a false positive result so that it cannot be a guaranteed method of confirming the presence of contraband.
 
 ## Background
 


beck-thompson:
I think this is a good solution to the shuttle bombing issue. Feels much better than just making explorations not work or something!

BramvanZijp:
What if the handheld one is a HoS steal objective item?

PotentiallyTom:
Having false positives wouldn't add anything, it just means anyone being flagged would be forced to walk through again. Having a plan involving no contraband, but but with clear bad-guy items be foiled by a radom false positive would be lame.

EthanQix:
I have a few suggestions about the design/balance of this device.

- The SCD should be located in the Security department. Its mechanics can be used beyond a mere evac checkpoint, and Security is where you'd want to be to conduct a thorough search during the shift. 
- There should only be one SCD on the station. This would induce a cost for a secoff wanting to conduct a more than just a random search, namely the time to bring the searchee to Security for the thorough check. That would also give the searchee a chance to book it if they do have sus shit on them, at the cost of their cover being blown.
- According to the previous point, I don't think there should be a portable version of the SCD. HoS would be all too happy to randomly check crew in the corridors.
- The SCD should be unanchorable so sec can setup a proper checkpoint at evac when the time comes, or anywhere they deem necessary at the time.
- There should be no way of building extra SCDs. Otherwise sec will be all too happy to put the whole station into 1984-style lockdown with a detector in every corridor. This would also make the SCD a valuable, if risky, target to destroy for traitors.
- Speaking of which, the SCD or the SCD board should be a high risk theif/traitor objective.
- Emagging the SCD should cause it to give random results. Much messier to deal with than a few false positives/negatives, as the machine would have to be deconstructed and rebuilt (especially bad with time constraints at an evac checkpoint).
- And last but not least, I'm not sure the SCD should detect anything beyond Syndicate items? Department-restricted items themselves are technically not illegal, and the machine would have to be smart enough (read: more coding required) to check against the searchee's ID. On the other hand, it would be funny to see the captain trigger all the beeps because someone stole their ID...

Aisu9:
this will not do anything regarding shuttle bombing, because shuttle bombing is in essence using uplink aboard shuttle, spawn explosive and trigger them. unless there is a plan to make uplink unuseable when shuttle reached the station.

you should also consider that this was introduced on TG, and everyone hates it. ain't talking about the handheld version that was even more hated. volume should ALSO not deafen people just to be sure. had enough of the contraband scanner gate SCREAMING in my ears because there is no options nor the volume itself was tuned down.

Killerqu00:
> this will not do anything regarding shuttle bombing, because shuttle bombing is in essence using uplink aboard shuttle, spawn explosive and trigger them. unless there is a plan to make uplink unuseable when shuttle reached the station.
> 
> you should also consider that this was introduced on TG, and everyone hates it. ain't talking about the handheld version that was even more hated. volume should ALSO not deafen people just to be sure. had enough of the contraband scanner gate SCREAMING in my ears because there is no options nor the volume itself was tuned down.

that is the plan (to lock uplinks)

Mixelz:
> Having false positives wouldn't add anything, it just means anyone being flagged would be forced to walk through again. Having a plan involving no contraband, but but with clear bad-guy items be foiled by a radom false positive would be lame.

It keeps the device from being able to be power gamed by introducing a factor of unreliability, one at the expense of security of needing to check someone that does not have anything and then purge their own record of the false accusation.

There is the possibility of it being a false negative instead, but I feel the onus should be on the antag to work with a bad dice roll rather than the station as a whole for one false negative allowing the shuttle bomber to slip in undetected.

src/en/space-station-14/departments/security/proposals/StationaryContrabandDetector.md
```diff
@@ -0,0 +1,62 @@
...
+> This would be a hard nerf to Storage Implants.
+
+I think it would make storage implants easier to detect, but it is an opportunity for more adaptive gameplay to emerge instead of using the implant as risk free storage.
+A very simple example could be holding major contraband in the implant while carrying minor contraband in a backpack to act as the scape goat.
+A competent Security could then ask for a quick recheck, sending the traitor into a panic for how to resolve this situation with their contraband still in hand.
+Effectively, we could roleplay as TSA agents and that itself sounds like it would lead to more fun situations than what we currently have available.
+> This would remove evac bombs from being possible.
+
+It would not since bypassing the detector is planned as part of its core intent.
+It WOULD make it much more difficult to achieve for the DaGD objective type, but it is already a trivial objective to accomplish and would be a massive QoL improvement for the Evac portion of a round.
```
Killerqu00, line 42: This is not what "QoL" means, though? It's just going to mean that evac bombing will be harder to do than:
- Go evac
- Buy bomb from uplink
- Activate and throw bomb

src/en/space-station-14/departments/security/proposals/StationaryContrabandDetector.md
```diff
@@ -0,0 +1,62 @@
...
+
+Considering its intended restrictions, I do not see many scenarios beyond Evac Bombing that this would be a major change for.
+It would be specifically alerting the behavior in regards to casually carrying contraband, which I feel is something that has not yet been dealt with in any direct capacity.
+It would also encourage more RP scenarios as mentioned above, which may not inherently be a buff/nerf to either faction depending on how the "Meta" evolves from it.
+
+## Potential Follow Ups/ Other Notable PRS
+Some potential follow up PRs to this device have been discussed, which will be noted below:
+
+### Locking Uplink when Evac arrives [#34359](https://github.com/space-wizards/space-station-14/pull/34359)
+This would be a very appreciated PR to ensure its intent of evac protection is upheld, without this it would be trivial to bypass it for Evac bombing purposes.
```
Killerqu00, line 58: That PR is basically required, should probably tell that in the beginning of the doc

src/en/space-station-14/departments/security/proposals/StationaryContrabandDetector.md
```diff
@@ -0,0 +1,62 @@
...
+### Allow the device to be EMAGGED so it always triggers a false positive.
+This will not be part of the initial implementation unless a maintainer would approve of it due to the EMAG restriction/freeze.
```
Killerqu00, line 60: I don't think any emagging interactions should be added at first. If I make a switch on/off that is access-locked, EMAG will already be useful here by allowing to bust the access lock and turn off the scanner.

src/en/space-station-14/departments/security/proposals/StationaryContrabandDetector.md
```diff
@@ -0,0 +1,62 @@
...
+### Create a handheld variant for individuals to hold in their hands.
+This would allow it to be more flexible in its use, though it should probably come with even harsher restrictions such as being unlocked with late tier research or being part of the armory.
```
Killerqu00, line 62: After some consideration, I don't think a handheld scanner is a good idea: it will just get used by security to check people randomly, no matter how you spin it.

src/en/space-station-14/departments/security/proposals/StationaryContrabandDetector.md
```diff
@@ -0,0 +1,62 @@
...
+
+A contraband detector, something that is a staple of real world security operations, would be a method to help alleviate these problems, add another layer of protection that traitors will need to bypass and would be a mechanical method of being able to detect if there is contraband in play while allowing sec to search a volume of people larger than themselves without grinding an area to a halt.
+
+Additionally, it can be used as a method to more reliably indicate the presence of a Storage Implant (I.E. Someone consistently setting one off despite having an empty inventory.), which has been a point of contention as Meta Knowledge due to the limited avenues of knowing when one is in play.
+
+## Design Intentions
+
+The intents of this implementation on when, where and why it should be used are as follows:
+1. As a starting point, it should trigger on any contraband outside of its owners department or always if it is syndicate contraband. An item in a dragged container cannot be attributed to a specific owner/holder and thus would always set it off without an owner being attached to it.
```
Killerqu00, line 27: Maybe make it ignore department-specific items completely...? The only "legal" bombs are security nades and salvage charges, which are not that destructive i think

Killerqu00 (commented):
hi i work on this thing

PotentiallyTom:
> It keeps the device from being able to be power gamed by introducing a factor of unreliability, one at the expense of security of needing to check someone that does not have anything and then purge their own record of the false accusation.

If this were implemented, then anyone knowingly carrying contraband has very little reason walk through it. So you end up with the majority of alarms being false positives. The majority of these would be solved instantly by asking whoever it was to walk through again, and the rest would just turn into `:s ward remove x from wanted please`. If the majority of alarms are false positives then people stop caring when it goes off, and we're back to square 1.

> Additionally, it can be used as a method to more reliably indicate the presence of a Storage Implant (I.E. Someone consistently setting one off despite having an empty inventory.), which has been a point of contention as Meta Knowledge due to the limited avenues of knowing when one is in play.

Adding any false positives completely blocks this as an option. Metashield on implants is broken when "experiencing a situation where absolutely no other explanation is possible". It's possible that the contraband scanner just keeps reading false positive.

Killerqu00:
> If this were implemented, then anyone knowingly carrying contraband has very little reason walk through it. So you end up with the majority of alarms being false positives. The majority of these would be solved instantly by asking whoever it was to walk through again, and the rest would just turn into `:s ward remove x from wanted please`. If the majority of alarms are false positives then people stop caring when it goes off, and we're back to square 1.

That assumes that the scanner is somewhere in a public spot. It is supposed to be either on the evac shuttle or in security for checking incoming prisoners. In the first case, you mostly don't have access to crimerecords - you just have the wanted list and search those marked by a scanner. In the second case, it just acts as QoL for the process of taking a prisoner in.

> Adding any false positives completely blocks this as an option. Metashield on implants is broken when "experiencing a situation where absolutely no other explanation is possible". It's possible that the contraband scanner just keeps reading false positive.

In that case, you can also rule out that due to bluespace existing, the item could have bluespaced somewhere out of the person's hands. Besides, there is a PR that makes implant checking an actually good mechanic with risk to security - I think that should be mentioned in the doc.

EthanQix:
> If this were implemented, then anyone knowingly carrying contraband has very little reason walk through it.

People wouldn't just "walk through it", it would be used by sec for a thorough search / checkpoint.

Pronana:
> Though speaking of, detectors giving false-positives WOULD pair really well with [#32136](https://github.com/space-wizards/space-station-14/pull/32136), it would give some risk and make secoffs think more carefully about checking for storage implants on everyone that trips a detector

They will already give enough non-antag positives when salvage and tiders walk through evac. I don't think we should add a random chance for them to have security search someone who doesn't have contraband, even worse if they then try to storage implant check them if that PR gets merged.

PotentiallyTom:
> That assumes that the scanner is somewhere in a public spot. It is supposed to be either on the evac shuttle or in security for checking incoming prisoners. In the first case, you mostly don't have access to crimerecords - you just have the wanted list and search those marked by a scanner. In the second case, it just acts as QoL for the process of taking a prisoner in.

No matter what, you're going to completely search anyone being brought into sec because they might have the tools to break out, so the only thing having a scanner in sec does is check storage implants, what does having false positives do for that except encourage implant checking? 

> People wouldn't just "walk through it", it would be used by sec for a thorough search / checkpoint.

Let's say I'm a syndie and I know sec are going to be checking people on evac. I'm not going to walk through the scanner with contraband on me. Whether that means I dump my contra before getting on the shuttle or find a way to avoid getting scanned doesn't matter. The scanner has done its job of stopping me from bringing a bomb onto evac. If I do get flagged, it's going to be a false positive because I have nothing on me. Everyone that gets flagged is going to be a false positive or a new player. These false positives are gonna be sorted out in 5 seconds by asking the person to be scanned again. All the false positive does is add 5 seconds to processing anyone and give innocent people random marks for the evac ride. You don't want people to stop trusting the scanner because then you get people ignoring it.

PotentiallyTom:
Unrelated to false positives but having it detect major and minor contraband makes it almost impossible to complete any steal objectives and escape. You would basically be forced to try to steal it on the shuttle, which is the opposite of what this PR is trying to do. It would probably need to be syndie contra only or have a blacklist for steal objectives.

Killerqu00:
> Unrelated to false positives but having it detect major and minor contraband makes it almost impossible to complete any steal objectives and escape. You would basically be forced to try to steal it on the shuttle, which is the opposite of what this PR is trying to do. It would probably need to be syndie contra only or have a blacklist for steal objectives.

There are many ways to bypass the scanner, such as:
- Board the shuttle not via main entrances
- Use any sort of technique to hide items (plushies or implants, no one is checking you for implants at the evac). You will be tagged, but it will be most likely deemed as a false positive
- Use a stolen ID/EMAG to disable the scanner manually

PotentiallyTom:
> There are many ways to bypass the scanner

> Board the shuttle not via main entrances

Then it's going to do very little to stop shuttle bombing, which is the first goal of the proposal. If it's going to be possible for ~3 syndies to bypass it every round for a steal objective, then it's going to be possible for 1 of them to also shuttle bomb. 

> Use any sort of technique to hide items (plushies or implants, no one is checking you for implants at the evac). You will be tagged, but it will be most likely deemed as a false positive.

"Ok you've been flagged as positive 3 times, we're going to be keeping you in cuffs until evac arrives at cc"

> Use a stolen ID/EMAG to disable the scanner manually

The only time that would work is if it checked for IDs in your inventory as well as the main ID you're wearing, which seems pretty arbitrary, otherwise sec would just instantly get you as soon as you changed ID.

Mixelz:
Love this thorough idea bucket for the machine! Thank you for taking the time to type all this out, I'll try to go through each piece here and then summarize my thoughts as a whole.

> * The SCD should be located in the Security department. Its mechanics can be used beyond a mere evac checkpoint, and Security is where you'd want to be to conduct a thorough search during the shift.
> * There should only be one SCD on the station. This would induce a cost for a secoff wanting to conduct a more than just a random search, namely the time to bring the searchee to Security for the thorough check. That would also give the searchee a chance to book it if they do have sus shit on them, at the cost of their cover being blown.

I believe using the SCD already comes with a significant operating cost, that being needing to plant an officer by the machine in order to actually manually search and detain any suspects. The machine itself cannot tell us what item set it off, only that something DID set it off. It is on the responsibility of the accompanying officer to follow through on that flagged result to actually find the source of the problem. Limiting it to one machine in sec would most likely ensure the machine is only used as a once over for confirming storage implants, which is not a design intention of the machine.


> * According to the previous point, I don't think there should be a portable version of the SCD. HoS would be all too happy to randomly check crew in the corridors.

I agree that random HoS maints checks would bad and should be discouraged. Perhaps the portable version could be more likely to give a bad result or have a long Do After, but that version will need it's own PR and balance discussion separately from this one.

> * The SCD should be unanchorable so sec can setup a proper checkpoint at evac when the time comes, or anywhere they deem necessary at the time.
> * There should be no way of building extra SCDs. Otherwise sec will be all too happy to put the whole station into 1984-style lockdown with a detector in every corridor. This would also make the SCD a valuable, if risky, target to destroy for traitors.

I very much like the idea of Sec needing to set up checkpoints for the device ahead of time, it should not be something that can be slapped down and forgotten about. Given the high value of this machine I would struggle to think of a scenario of destruction that wouldn't just end up as throwing a c4 bundle ontop a group of sec offs and running away. If instead we devalue the machine it would be much more likely to be left alone, and then therefore sabotaged through means that aren't just blowing it up.

> * Speaking of which, the SCD or the SCD board should be a high risk theif/traitor objective.

Very fun idea! I think this would work well on the portable version, though I also worry about having too many steal objs for Sec equipment.

> * Emagging the SCD should cause it to give random results. Much messier to deal with than a few false positives/negatives, as the machine would have to be deconstructed and rebuilt (especially bad with time constraints at an evac checkpoint).

This would be much harder to notice than always giving a positive result. I think that determining if the device is Emagged shouldn't be a guessing game and should instead be something obvious that will require time and effort to undo, much like emagging a door and needing to take the time to unbolt it and reset it's permissions.

> * And last but not least, I'm not sure the SCD should detect anything beyond Syndicate items? Department-restricted items themselves are technically not illegal, and the machine would have to be smart enough (read: more coding required) to check against the searchee's ID. On the other hand, it would be funny to see the captain trigger all the beeps because someone stole their ID...

The idea of it scanning everything by default was to ensure that it had value outside of Syndicate rounds, such as Revolutionaries, Particularly Ornery Tiders or Theifs needing to sneak past with their objective in tow. If it ends up being too restrictive for traitors it can be reigned into something such as only including Command Items or only High Tier contraband.

In addition, Tying it to IDs is a method to ensure that you cannot effectively "hide in a crowd", assuming an evac scenario where you have a servers worth of people all come through at once. That being said, I am all on board for having the cap get flagged by security due to ID shenanigans because adding potential misinformation chaos should be a valid method of getting past it as a traitor.

Based on the overall discussion, we'll probably want to keep the device as something that only exists outside of Security at a designated/manned checkpoint and emphasize that it should be tremendously difficult to produce additional ones, if we allow them to be craftable at all.

src/en/space-station-14/departments/security/proposals/StationaryContrabandDetector.md
Mixelz: QoL in the sense that dealing with an Evac bomb is a negative quality of the end of round play.
I will update the wording regardless. 👍

src/en/space-station-14/departments/security/proposals/StationaryContrabandDetector.md
Mixelz: Yep, that is why EMAG as a baseline effect is in the follow up section.
I'll adjust wording so the idea of an EMAG effect itself is a follow up, with this specific example being just *an* example of how one would achieve this.

Mixelz:
> > Though speaking of, detectors giving false-positives WOULD pair really well with [#32136](https://github.com/space-wizards/space-station-14/pull/32136), it would give some risk and make secoffs think more carefully about checking for storage implants on everyone that trips a detector
> 
> They will already give enough non-antag positives when salvage and tiders walk through evac. I don't think we should add a random chance for them to have security search someone who doesn't have contraband, even worse if they then try to storage implant check them if that PR gets merged.

The alternative would be having this thing be a 100% confirmation of the presense of contraband itself or of a storage implant, which I think would solidly push the item as being too powerful. 
I think that allowing a chance for security to be wrong is the better resolution here, though through a false positive or false negative is still undecided.

src/en/space-station-14/departments/security/proposals/StationaryContrabandDetector.md
Mixelz: Possibly, but we can come to that decision if the stationary variant gets through and if it is deemed as something worth  considering to add.

Mixelz:
> If this were implemented, then anyone knowingly carrying contraband has very little reason walk through it. So you end up with the majority of alarms being false positives. The majority of these would be solved instantly by asking whoever it was to walk through again, and the rest would just turn into :s ward remove x from wanted please. If the majority of alarms are false positives then people stop caring when it goes off, and we're back to square 1.

That is why it should be a low chance for a false result as suggested. I would agree with this point if it was going to be guarenteed effect but the specific amount for this to feel fair will need to be tested and itterated upon to figure out a healthy balance. Figuring out that balanced value is something that is outside the scope of any design document, only the general idea of it being a "low" chance is required as a baseline.

src/en/space-station-14/departments/security/proposals/StationaryContrabandDetector.md
Mixelz: Agreed, giving it is own section higher up in the document.

src/en/space-station-14/departments/security/proposals/StationaryContrabandDetector.md
Mixelz: Changing the wording to specify all contraband besides department specific contraband. This will probably need to be played with but the intent is that it cannot differentiate between a single wooden baseball bat and a fully stocked C4 Bundle.

ilovehans10:
One thing that would make it harder to move would be making it a multiblock with three separate parts that all need to be anchored. The different parts could be something like control computer, scanner, and power modulator. It could also require one of it's pieces to be on top of MV or HV wire directly, so that you need a t-ray scanner or crowbar to get placement correct. Granted none of these would fully stop sec from moving it, but if only one sec officer is moving it a long distance it would provide a tider with a fair bit of time to walk over and hack one of the pieces not yet moved.

I really like false positives, even though they do raise some interesting implementation questions. One idea to weaken the "just walk back through" meta would be to have the scanner keep a list of recent false positives it's done, and if one of those players walks through again, they have a high chance of getting another false positive. This would mean that sometimes walking back through would work, but other times it would just set off the scanner again and make sec redouble their efforts. This list could then have items removed from it after a set period of time so players aren't cursed with false positives for all time.

Another idea would be to have certain items be targets of "scanner miscalibration". This miscalibration would be applied round start, and would choose a random item from a list of rareish items to always flag. For example, if mousetraps are selected at round start as the miscalibration item, then if anyone gets scanned with a mousetrap in their inventory they will scan as positive the same as if they had contraband. Some other example items could be things like bar shakers, jaws of life, kitchen knife, or any other item that close to only 1% of players will have on them. The idea in this paragraph is definitely half-baked, but I just wanted it out in the world for feedback/ideas.

I think an interesting way to do counterplay with this would be to have the radio jammer interfere with this in some way, but I am unsure whether it would be best to have the jammer block the scanner, or make it give a false positive at range. I feel like if the jammer is used to block the scanner it would have to only work for a couple of seconds, after which the machine goes into some sort of error mode and notifies its operator that it is being jammed. The idea in this paragraph is definitely half-baked, but I just wanted it out in the world for feedback/ideas.

Overall, I love this idea, and wish the best of luck on getting it approved and then implemented.

lorahart2:
What if it spurts out a picture of their character like used their id picture of them in Grays with a single outline of the most dangerous item on them either a gun or a bomb

0x6273:
I think this should be limited to only detecting explosives, and should only exist on the evac shuttle.

lorahart2:
What of it detects both unspent TC and explosives meaning you should be incentivized to preplan

Smugman:
I personally really do not like this, especially in regards to steal objectives, and ESPECIALLY with a portable version.
I think this structure is a bandaid solution that would outsource a lot of the detective work of security.
If this is yet another PR mainly conceptualised to deal with shuttlebombing, then it should likely be constrained to shuttlebombing instead of detecting every kind of antag activity at once. 

> * Shuttle is a way easier way to achieve almost any objective due to how small it is. There should be something that makes it harder to just wait until the shuttle to do objectives, and this scanner aims to be one of these things

Yes, but it is also significantly harder to go unseen when acomplishing objectives. The EVAC shuttle is incredibly chaotic, and security is appropriatly short tempered there. In most cases any incident on the EVAC shuttle is addressed by simply cuffing them and waiting to "deal with them at CC" and/or lethalling them.

By this same logic, it is far too easy for Security to accomplish their goal of apprehending traitors on the shuttle, when the round has already concluded.

> * There are many ways to bypass the scanner, such as:

That is not the point, the fact is that this is yet another complication to steal objectives which is unneeded. If you manage to steal an object unseen, and manage to avoid random searches or hide your contraband cleverly, I do not think that a random machine at roundend should immediatly bust you at no cost or effort to Security.

Searches can be avoided by clever players in two ways
1. Dont get searched, Simply dont be suspicious, or avoid security officers during their "random search" period. If the CMOs hypospray got stolen, security is most likely to search people close to medical, for example. 
2. Hide your contraband. Maybe you have a storage implant to hide your viper in. maybe you stuffed the Emergency Orders into your Survival box, or your zipped up labcoat, or another place that not every security officer knows to check. 

The Contradetector circumvents both of these, especially the storage implant (Free metashield break exploit, 2655, unpatched)
You are forced to walk through it, and no matter how cleverly you hide it, or even if you spent your limited TC on anti-search items, it will detect them.

> "Ok you've been flagged as positive 3 times, we're going to be keeping you in cuffs until evac arrives at cc"
This is very accurate to how this is going to go. 

> If this were implemented, then anyone knowingly carrying contraband has very little reason walk through it

Most Evac shuttles would likely be set up to not give you a choice, or security would funnel everyone through the one entrance with a Scanner.

I feel this is very overcorrective and will be relentlessly powergamed, no matter how high the rate of false positives or negatives is.

Killerqu00:
> That is not the point, the fact is that this is yet another complication to steal objectives which is unneeded. If you manage to steal an object unseen, and manage to avoid random searches or hide your contraband cleverly, I do not think that a random machine at roundend should immediatly bust you at no cost or effort to Security.

"it should trigger on any contraband that is not department specific"
emergency orders (which don't exist anymore btw) are department-specific (command), as well as hypospray. this does not affect steal objectives in any way, since steal objectives are either department restricted or not even contraband

> The Contradetector circumvents both of these, especially the storage implant (Free metashield break exploit, 2655, unpatched) You are forced to walk through it, and no matter how cleverly you hide it, or even if you spent your limited TC on anti-search items, it will detect them.

1. The detector is supposed to be mapped on evacshuttles and in brig. If you're in brig, the odds are you're being searched already. If you come to evac with syndicate contraband, consider either not doing that or at least being somewhat smart about it.
2. If you're getting searched in the field (not in security), then you're not being scanned anyway, meaning that strat is working. Otherwise, the officer won't be able to find the stuff anyway, since they don't know what to check, specifically. A case could be made for repeated scans, but I'm currently working on that flaw in design

PR #382, opened 2025-01-15
Add review exemption for minor bugfixes

SlamBamActionman:
As discussed internally within the maintainer team, this exemption allows a single maintainer to merge/close minor bugfixes, bypassing the "2 Maintainer" requirement.


ArtisticRoomba:
> What happens if the PR has a content bundled with bug fix, or art asset/respriting with bug fix

as per PR guidelines these should be atomized into separate PRs for easier review and managing already

SlamBamActionman:
Exemptions are also optional; while there may be edgecases, we put some trust into maintainers exercising common sense whether an exemption is warranted. And with the biweekly release review we have an additional safety net to catch things before release.

Errant-4:
> And with the biweekly release review we have an additional safety net to catch things before release.

Although, now that I think about it, a PR that accidentally slipped through as a Bugfix might likewise slip through the feature list, which does not include bugfixes. We will just have to be vigilant about the list, I guess

PR #383, opened 2025-01-17
New Cloning System Proposal

Velken:
Proposal Doc for new cloning system
Based on the fact that many players don't like being turned into borgs

I haven't written a design document in a long while, and its my first time writing one for SS14, hopefully I did everything right.


Ratonero500:
I really like the overall idea and how its balanced, but i think it would be more balanced if whatever machine makes the new bodies would take 3-5 mins to make a body and require a lot of power(sci levels of power), if the machine loses power in the process, its progress would decay at half the rate it increases, so if its 2 mins in, it would take 4 mins to degrade to 0, this would make cloning an intensive process and make interesting situations where you must choose who revive first.

This would be a big trade off for cloning, since if you want to be cloned you must wait the time the machine takes to make a new body instead of the short time that it takes to assemble a cyborg.

PR #385, opened 2025-01-18
Add note about using `savegrid` for CentComm

minus1over12:
[CentComm should be saved as a grid.](https://github.com/space-wizards/space-station-14/pull/34475#issuecomment-2597684314) This clarifies that for new contributors.


Partmedia:
Thanks!

PR #394, opened 2025-02-05
Crawling Proposal

Plykiya:
The follow-up to https://github.com/space-wizards/space-station-14/issues/34894

I don't have all the details fully ironed out yet, and I don't have much experience with how crawling is actually implemented in SS13 or SS14 forks, so it would be nice if people could discuss and review this. I laid out what I believe are the most common implementation points, but please let me know if I missed anything.


walksanatora:
uhh did you intend to delete/close this PR?

PR #395, opened 2025-02-07
"Move Circuit Imprinters to Engineering" Proposal

Compilatron144:
Proposal doc for moving the circuit imprinter from science to engineering.

Gives engineering something to do - thus extremely controversial.


walksanatora:
I remembered my "remix station" idea. which had many stupid ideas. this was one of them. the others were
- the primary to get between departments is via mail and diposal tubes
- robotics is located next the AI core (robotics on the sat. not AI core in sci)
- atmos gets waste proucts and has to recycle them for distro (you get no2, and co2 miners. but as a bonus you get free plasma miner for your troubles)
- 2 singulo chambers (like the 2nd one is gonna get used)

PR #396, opened 2025-02-08
Proposal- Cephalopod Race

Waluigitime4lyfe:
Proposal Doc for cephalopods as a playable race in the character customize menu.


IamVelcroboy:
Cephalopeople!

PR #406, opened 2025-02-20
Botany proposal

MilonPL:
This proposal aims to integrate botany into the kitchen/chef gameplay loop rather than making it a separate role. Botany doesn’t have much direction, and chef gameplay has its own issues. Food quality doesn’t really matter, and there’s almost always way too much food on the station. 

This doesn’t fix that, but the goal here is to give chefs something more to do. Instead of botany being its own isolated thing, it’s getting folded into the kitchen as a side activity. Chefs will have a small hydroponics setup to grow ingredients, with a more streamlined system that keeps things engaging without taking over the role. 

Chemistry could have their own hydroponic setups, or a chef could handle plants during slower hours and take orders from med. This also helps fix issues like botany not growing ingredients for the kitchen or there being no botanists at all.

### DIFF
--- a/src/en/space-station-14/departments/service/proposals/botany-proposal.md
+++ b/src/en/space-station-14/departments/service/proposals/botany-proposal.md
@@ -1,0 +1,106 @@
+# Botany 
+
+| Designers | Implemented | GitHub Links |
+|---|---|---|
+| MilonPL | :x: No | TBD |
+
+## Overview
+
+This design document proposes the removal of botany as a standalone sub-department and its integration into the kitchen gameplay loop. The goal is to streamline plant growing mechanics while making them more meaningful as part of food preparation, turning hydroponics into a side activity for chefs rather than a separate role.
+
+## Background
+
+Currently, botany suffers from several issues:
+- Heavily RNG-based gameplay with little player agency
+- Limited interaction with other departments except kitchen and medical
+- Lack of meaningful progression or goals
+- Redundant role that could be better served as part of existing gameplay loops
+
+By integrating botany into the kitchen, we can:
+- Give chefs more varied gameplay during slow periods
+- Maintain the essential function of growing ingredients
+- Create a more focused and purposeful growing system
+
+## Features to be Added
+
+### Kitchen Garden
+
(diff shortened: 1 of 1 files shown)


keronshb:
I personally have a different view for Botany. Some of it is from unused SS13 elements.

On SS13 there's steelcaps which allow you to make steelwood and durathread, which allows you to make durathread.
Though Botany really shouldn't be just about those 2 items, but they could help improve the station.

Wood > Steelwood could be an interesting progression. On Cit we had this as a prerequisite for training katanas/bokken.
Durathread was a huge direct upgrade to regular storage (not as much as a BOH but a significant increase) and it allowed for different upgrades to cloth items.

This does have some slight overlap with sci by improving the station, but it does take some of the workload off of sci and puts it to botany.

Another idea I had was like, add similar plants to steelwood and durathread. And botany could submit plants to sci to deconstruct/research so they could in turn make mutations easier or help unlock things.

HTMLSystem:
I disagree with this. As someone who has tried to do botany/chems and be a chef/botanist it's a lot of multitasking. If you're not on your botanist game you're not gonna make any real progress. Food is going to be slow to make and so would chems. Having a dedicated role focused on making and mutating plants is the best way to go about it so every department can get the best plants they need for their work.

HTMLSystem:
Also, with the problems you stated with botany (heavy RNG, no real progression (disagree)) this wouldn't fix anything but force chefs into the bitany gameplay loop that they might not want to take apart of.

Ratonero500:
I feel like there is a niche of players that like the whole complex RNG based nature of botany as is, and for that they play botany, this feels like destroying a really good niche only for simplification.
i feel like other solutions to this problem with chefs being not that necessary  would be better, one example could be the addition of moods and food affecting mood.

NoElkaTheGod:
"Remove botany" ahh proposal.

Yes, botany has some problems, but, as stated above, this PR is just oversimplificating things that won't even work that way. Growing plants is not "add the seeds, come back 10 minutes later and get the results", it is a complex process that requires constant monitoring, especialy so if mutagen is invloved.

PR #410, opened 2025-02-28
Scar's AI Proposal

ScarKy0:
Yet another AI proposal document.

This one is based of Aeshus' which is based on Saphire's which is based on Moony's

The general idea is a resource oriented AI that is integrated with the station and silicons.


Djungelskog2:
My issue with these "minimax AI" docs is that they often take away from the AI's job of interacting with crew normally and make it even more overwhelming than it already is. I think it also adds unnecessary friction between crew and the AI to the point it would be hated more than "validhunting AI" (it's just following it's laws). The AI shouldn't be all powerful in what it can do directly and if it were, it should do so through the form of a remote control chassis or other hands on means. I don't think the AI should have so much control over it's hardware, mayhaps require most sensitive systems to have hands-on actions (done through your shell, borgs or people)? The AI should be able to do it's job effectively very easily and should have it's core tools such as door control, atmospheric control of individual rooms, turret control and cameras for free (Or essentially free to do even on a wide scale). I think the AI core being a secure area that people shouldn't go into often is a perfectly sound and good design choice. I would much prefer the malf AI be reworked to fit within a crew-sided toolkit and have to work around that (considering that's what it's made for) where it has access to all the tools it needs at it's disposal, and simply has to try harder to remain undercover; if the ai hacking APCs reveals it as malf, get rid of the hacking APCs part, dont make the normal AI hack APCs also. This could be done through a completely alternative task system. As of now the AI is not at all "too powerful" and I don't imagine it would be when it gets full functionality with more machines and interaction with silicons.

SlimmSlamm:
this nerfs the AI so hard and just makes it a pain in the ass to play. i dont see myself wanting to ever touch AI again if it had to rely on processing power

Ratonero500:
I feel that the only limitation to current ai should be that if more than 10-15 bolted electrified doors are active at the same time,then it should consume compute power.

Camera,radio and closing opening doors consumin compute power feels annoying.

ASTRALIXIX:
For some reason, it seems to me that creating a new AI should be more difficult than the usual insertion of a positronic brain into an Intellicard.

walksanatora:
on the topic of processing power. I think all *Current* abilties AI have should remain free. but any more advanced abilties. or bulk applications of current abilitie should be given resource cost. eg: read how paradise station does it. https://www.paradisestation.org/wiki/index.php?title=AI#Programs all advanced abiltites need memory and time or nanites to use.

Co0kie-GG:
https://github.com/Monkestation/Monkestation2.0/pull/11862 In fact, a similar AI concept already exists on some builds in 13 (In fact, only the former Yog) And it could well be improved, as it seems to me...

ScarKy0:
ohhh cool stuff ill have a look

EthanQix:
Do we have a document anywhere summarizing in a couple sentences what the AI role is supposed to be and do?

PR #417, opened 2025-03-09
Blood Cult Proposal

kbarkevich:
This proposal aims to integrate the Blood Cult antagonist (originally from Space Station 13) into Space Station 14.

### DIFF
--- a/src/en/space-station-14/round-flow/proposals/blood-cult.md
+++ b/src/en/space-station-14/round-flow/proposals/blood-cult.md
@@ -1,0 +1,169 @@
+# Blood Cult: Major Team Conversion
+
+| Designers | Implemented | GitHub Links |
+|---|---|---|
+| Ze SpyRo | :x: No | TBD |
+
+## Overview
+
+This document describes a new Team Antagonist, the Blood Cult. The Blood Cult is initially composed of two or three non-command, non-security, and non-chaplain crewmembers who paint runes on the floor and use blood magic to kill, kidnap, or convert their fellow crewmates into the cult. Their goal is to sacrifice up to two specific high-ranking crewmembers in order to prepare the final ritual that will allow them to summon Nar'Sie, winning the round.
+
+## Background
+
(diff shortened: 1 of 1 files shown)


src/en/space-station-14/round-flow/proposals/blood-cult.md
```diff
@@ -0,0 +1,169 @@
...
+- **Fun to lose against (or as)** - Losing to a cultist is unlikely to remove a player from the round. If converted, they themselves will become a cultist. If sacrificed, the soul stone that drops out of the ritual may be used to summon constructs that the victim player can control, keeping them engaged in the round. Losing *as* a Blood Cultist is likewise unlikely to remove the player from the round, as the station's primary method of countering a cult is deconversion using holy water. The Chaplain is the clearest counter to the Blood Cult, being highly resistant to their weapons/abilities as well as the station's sole producer of holy water. However, they are not completely safe; Blood Cultists may use more advanced strategy and more traditional (non-magical) means to assassinate a troublesome Chaplain.
```
beck-thompson, line 146: Like we talked about in discord, it might be a good idea to allow holy water to be purchased in cargo in case the chap dies, is ssd, or not even on station!

beck-thompson (commented):
My other suggestion is to allow mindshields to be removed. It's a thing in funky but I'm pretty sure it isn't a thing upstream! It seems important so crew who get the mindshields aren't necessarily RRd.

ScarKy0:
> My other suggestion is to allow mindshields to be removed. It's a thing in funky but I'm pretty sure it isn't a thing upstream! It seems important so crew who get the mindshields aren't necessarily RRd.

Its actually a wanted feature, for revs as well
tbh it SHOULD be a small PR i hope

Pumkin69:
this is insanely cool.

chromiumboy:
Overall I like the proposal

I do agree with beck's comment though, that such an important mechanic like deconversion shouldn't require having a chaplain on staff. I don't mind the idea that chaplain offer some minor perks in fighting cultists though. For example, chaplains could be able to bless weapons so they deal more damage to cult constructs, or the ability to sense nearby cultists when the cult advances to the stage that their eyes start glowing

PR #419, opened 2025-03-13
Chemistry proposal

kosticia:
Small document that should make chemistry a little bit more interesting.


MarzieK:
By making almost everything randomized you're making it less fun, unorganized and straight up frustrating. No.

kosticia:
> By making almost everything randomized you're making it less fun,

Yes, I agree. Chemistry right now is so much fun. I make ipecac and laugh.
>unorganized

Sure, yes. In game about random all should be organized. I'm going to mix my favorite iodine 12 hours in a row, because its so much funny.
>and straight up frustrating.

"Today I played xenoarch and lost."

Dragonjspider:
This idea has some merit, I like the idea of a randomized chemical that a chemist can experiment to find when they have downtime, something for them to do. But having this system completely change the already well known chems, of which you have to make lots of in every round,... I can't imagine having to write down a single mix **(which may or may not need to reach a certain temperature)**, let ALONE every single medicine that you have to make EVERY round...

I can easily see rounds happening where chem wasn't able to even make Bicaradine just because they spent the last 40 minutes trying every mix at every ratio, only for it to need a high temperature to work...

If this isn't how it's supposed to work, then the design doc isn't clarifying that... Either way, the design seems undercooked.

For example:
- What will be the process of discovery? Just mixing until you randomly find what you want?
  - If not, will there be a randomly generated note in some way? Or from examine information?
- You never clarified if some reactions to make chems may need a temperature change, will this be dependant on the result or is it an independent requirement? (I.E, If I want to make Bic, will I have to worry about the ingredients needing to be hot, or will it stay the same?)
  - Will this also be the same for plasma?
- What about the Agent on nukie rounds? How do you expect them to work alone with this in mind?
- What about the syndicate specific chems? will they also need to worry about these changes?
  - If so, how would you change it to be reasonable?

No matter how you look at it, this won't just effect chemists, this will effect everyone, making a mechanic more involved doesn't necessarily make it a better experience...

TLDR;
Drop the change to ALL chems, make it a mechanic to some new or existing chems with unique effects (Perhaps instead it would modify how some chems act?), and I think it's an interesting idea. Otherwise, it might need a lot of consideration with how it will impact rounds that are... heavily impacted by medical's structure... (War-ops)

kosticia:
> I can easily see rounds happening where chem wasn't able to even make Bicaradine just because they spent the last 40 minutes trying every mix at every ratio, only for it to need a high temperature to work...

I can't say that chem with random reagents became too much harder. Chemist with that system just needs to found reagent that heals brute with its goggles, them probably increase its effect with others reagents.

> * What will be the process of discovery? Just mixing until you randomly find what you want?
>   
>   * If not, will there be a randomly generated note in some way? Or from examine information?

Partially yes, chemist should mix random reagents, and them examine them with a goggles, that will show effects of the reagent.

> * You never clarified if some reactions to make chems may need a temperature change, will this be dependant on the result or is it an independent requirement? (I.E, If I want to make Bic, will I have to worry about the ingredients needing to be hot, or will it stay the same?)
>   
>   * Will this also be the same for plasma as a catalyst?

No, random reagent reactions won't need to be heated or have a catalyst. Just two reagents, you mix it, and get new one.

> * What about the Agent on nukie rounds? How do you expect them to work alone with this in mind?

I think that nukie rounds with a random reagents is a little problem. And nuclear medicine probably should moved into uplink, fact that one guy from nukie team buys something from uplink instead of click on chemmaster not a big loss.

> * What about the syndicate specific chems? will they also need to worry about these changes?
>   
>   * If so, how would you change it to be reasonable?

Vestine and stimulator just will be stable reagents that will have useful for traitors reactions.

> No matter how you look at it, this won't just effect chemists, this will effect everyone, making a mechanic more involved doesn't necessarily make it a better experience...

As said later, I don't think that random reagents will make chemistry harder. Yes, there will be some new mechanics that need to be learned, but they can be putted into guidebook.

> If your intention is only to have chemistry be more chaotic in the same way science is, it's probably better suited to unique chems with unique effects, that way you keep the experimentation, while also keeping the already well documented and well known chems intact. This way you can still have that chaos, but none of the frustration of trying to make any of the necessary medical supplies that the CMO's been yelling about for the past hour.

Interesting idea, but I still think that there is only two ways for chem.
First one - leave everything as is.
Second one - random chem according to this design-doc.

FairlySadPanda:
Drive-through comment: future med on Wizden won't be as cripplingly addicted to chem to provide all healing, so there's totally room to explore chem in a more sci-like favour.

MozarteanChaos:
I feel like one of the problems here is that while this spices things up for experienced chem players... this will probably not be pleasant for them, or for the newbies. Veteran chem players will have to re-learn everything every round, and new players have to juggle learning chem's mechanics and figuring out - from scratch - which chemicals can be used as medicine.

This also doesn't bode well for doctors, IMO, since it's going to be harder for chem to share knowledge of which reagents are safe to mix, and impossible for doctors to remember this information from previous shifts. Currently, doctors can be told not to mix brutes, and retain that information - some will even learn that dylo and inap shouldn't be mixed. They won't be able to learn this if most chemicals and reactions are randomized.

You're also removing depth from chem at the same time, by turning most recipes into two-reagent, no-heating reactions. I don't think this is particularly helpful given the document's stated goal - yes, it makes the random recipes less painful, but if they can't exist without removing depth from the system they're meant to _add_ depth to, there may be a problem with your approach.

SlimmSlamm:
Yeah let's totally randomize it so people who are dead have to wait longer to play. Slop idea.

walksanatora:
> Interesting idea, but I still think that there is only two ways for chem.
> First one - leave everything as is.
> Second one - random chem according to this design-doc.

you are forgetting the 3rd option. keep existing chems. but let chemist "expierment" with mixing extra chemicals into existing chems for the chance at better or diffrent effects. thatway chemist dont need to waste over half the round expiermenting to get basic chems out. but also allows mid-end round chemist the ability to expierment and see what amazing effects they can find or create.

walksanatora:
or again. give chem some kind of "chem x" that when mixed with chems (mabey require heating) will "mutate" the chem
in a consistent chain. so you could get upgraded bic that somehow has cognizine effects. and mabey with enough "mutations" it could loop back into another chem (accidentally makes lexorin from carbon)

Dragonjspider:
> > I can easily see rounds happening where chem wasn't able to even make Bicaradine just because they spent the last 40 minutes trying every mix at every ratio, only for it to need a high temperature to work...
> 
> I can't say that chem with random reagents became too much harder. Chemist with that system just needs to found reagent that heals brute with its goggles, them probably increase its effect with others reagents.

So let me get this straight, chems at a base have effects, which can be made better by mixing other chems with different effects?
Does this mean different base chems might have different effects when combined with other base chems?

As an example, X heals 5 brute, Y heals 5 Burn, Z increases healing done by 3. (I'm going to assume Z will not be recursive* here)
So if I mix X and Z, X now heals 8 brute, right?
And I assume Y and Z would heal 8 burn as well.
So if you mix X, Y AND Z, then it heals brute and cold by 8?

If this is true, what's stopping chemists from making a superpowered Omnizine by mixing every chem with a positive effect?

And to add to that, what about chems with negative effects? I assume you might be able to neutralize those with a different chemical?

Also how would this affect ratios?

- [1] Oxygen, [1] Carbon, [1] Sugar = Inaprovaline [3]
- [1] Inaprovaline, [1] Carbon = Bicaradine [2]

Would it be 1-1 on every chem, or would some randomized chems have their own ratio modifier?
Or would modifying chems at a 1/2 ratio?

- [1] X, [1] Z = W [1]?

> > * What will be the process of discovery? Just mixing until you randomly find what you want?
> >   
> >   * If not, will there be a randomly generated note in some way? Or from examine information?
> 
> Partially yes, chemist should mix random reagents, and them examine them with a goggles, that will show effects of the reagent.

Alright, this makes sense, as much as I would love an excuse for chemists to experiment on monkeys, I can imagine that being time consuming.
I feel like this defeats the purpose of having all of the chems be randomized though...

> > * You never clarified if some reactions to make chems may need a temperature change, will this be dependant on the result or is it an independent requirement? (I.E, If I want to make Bic, will I have to worry about the ingredients needing to be hot, or will it stay the same?)
> >   
> >   * Will this also be the same for plasma as a catalyst?
> 
> No, random reagent reactions won't need to be heated or have a catalyst. Just two reagents, you mix it, and get new one.

Okay, this poses a new question: What about existing chemicals you can get from grinding down items?
I assume they will still exist as unique modifiers, but I can see this being complex coding-wise too...
(This goes back to the "negative effects" question I asked earlier.)

> > * What about the Agent on nukie rounds? How do you expect them to work alone with this in mind?
> 
> I think that nukie rounds with a random reagents is a little problem. And nuclear medicine probably should moved into uplink, fact that one guy from nukie team buys something from uplink instead of click on chemmaster not a big loss.

Might be stepping on eggshells there, having to buy chems from an uplink isn't a bad idea, sure, but that implies they'll have to spend TC, which then means you'll have to balance the prices around it, and likely increase the amount of TC the agent has. And  as a result, there's also the possibility that they might skip over chems entirely just so they can have a stronger offence. (probably a dumb idea on the nukies, but likely indirectly causing more Blitz-op rounds)
 
> > * What about the syndicate specific chems? will they also need to worry about these changes?
> >   
> >   * If so, how would you change it to be reasonable?
> 
> Vestine and stimulator just will be stable reagents that will have useful for traitors reactions.

Saw your update on the doc, makes sense, adds a lot of creativity to what you can do with them. I like it!

> > No matter how you look at it, this won't just effect chemists, this will effect everyone, making a mechanic more involved doesn't necessarily make it a better experience...
> 
> As said later, I don't think that random reagents will make chemistry harder. Yes, there will be some new mechanics that need to be learned, but they can be putted into guidebook.

I didn't say this to say it would make something harder, sorry if it seemed like I meant it that way. I just meant that the process might take more steps, and I'm asking these questions just to see to what extent these changes would effect the process of chemistry as a whole.

> > If your intention is only to have chemistry be more chaotic in the same way science is, it's probably better suited to unique chems with unique effects, that way you keep the experimentation, while also keeping the already well documented and well known chems intact. This way you can still have that chaos, but none of the frustration of trying to make any of the necessary medical supplies that the CMO's been yelling about for the past hour.
> 
> Interesting idea, but I still think that there is only two ways for chem. First one - leave everything as is. Second one - random chem according to this design-doc.

Now that I understand better what the doc is supposed to do, I do get it. But I have a very important question about the intention overall.

If chemists can figure out what a chemical does by examining it with chem goggles, then it just slows down the process doesn't it? I like the idea of some chems being modifiers to the resulting chems, sure. But changing ALL chems doesn't change much compared to adding new chems that modify the existing ones. Making bicaradine for example, just goes from:

1. Mix [1] Oxygen, [1] Carbon, [1] Sugar = Inaprovaline [3]
2. Mix [1] Inaprovaline, [1] Carbon = Bicaradine
3. Repeat until out of chems, go to cargo / See if botany has that miracle plant that is free bic.

To:
1. Look through the 10+ randomized chems you may or may not start with to find each chem that heals brute damage
    - (Here comes the "nope not that one, nope not that one, nope not that-")
    - (Didn't find what you wanted? Go to cargo [*1], gamble for it [*2], Return to step 1.)
2. (Optional) Find chems that modify it to be better
    - ("nope not that-")
    - (Didn't find modifiers good enough? Consider going to cargo[*1] or move on.)
4. Mix the (likely) three separate chems to make something similar to bicaradine

I see what you're trying to do, I really get it. But what's the reason behind making the entire of chemistry's process slower, when you can keep the experimentation for improving existing chems?

[1*] You mentioned in the doc that botany will have some use, but there wasn't any examples you gave, so I'm assuming cargo and botany are interchangeable here.
[2*] MFW botany gives me a chem that does the exact opposite I was looking for: (now they have to restart their own gambling mission to get me what I want)

src/en/space-station-14/departments/medical/proposals/chemistry-rework.md
```diff
@@ -0,0 +1,61 @@
...
+
+| Designers | Implemented | GitHub Links |
+|---|---|---|
+| kosticia  |    :x: No   |    To do     |
+
+## Overview
+Right now chemistry is the most boring sub-department. Every round – “silicon + nitrogen + potassium”, then “oxygen + plasma” and over, and over… Zero interaction with 99% of crew. This proposal contains potential solution of this problem. 
+
+## Random reagents
+Main idea of this proposal – make round start chem reagents and some reactions for base medical random-generated. How this will work? Round-start chem-dispenser will have reagents with random names and functional. Reactions for these reagents also will be random, for example, reagent 1 and reagent 2 can be mixed into powerful brute topical, reagent 2 and reagent 3 after mix will create explosion.
```
Dragonjspider, line 11: Do Reagents 1 and 2 have their own effects when ingested? Will chemical goggles show what reactions to expect?

What about Reagent 3? In this scenario, is that Reagent the reason the mix exploded, or is it the result of mixing Reagent  2 with 3?
(As in, if I mixed 1 with 3, would it still explode, or is that a unique interaction with 2?)

src/en/space-station-14/departments/medical/proposals/chemistry-rework.md
```diff
@@ -0,0 +1,61 @@
...
+
+### Random reagent reactions system
+All random reagent reactions should take only two reagents and not require heating or catalysts. Idea of hotplate - make some reactions chemistry available only. With new chem, this idea will be a little bit irrelevant, because better ask chemists for working recipe, than get blown up durring mixing.
```
Dragonjspider, line 14: Wording is a bit confusing here:

> Idea of hotplate - make some reactions chemistry available only
Took me a bit to understand you meant
"The idea of hotplate before this doc was to make some reactions only possible in the chemistry lab"

(Nothing is preventing a tider from microwaving chems to make something)

src/en/space-station-14/departments/medical/proposals/chemistry-rework.md
```diff
@@ -0,0 +1,61 @@
...
+After mixing, new reagent will have more powerfull effect from two mixed ones, and one new, passive.
```
Dragonjspider, line 15: As mentioned in my comment in the conversation area, what kind of examples of passives do you mean?
Could they change how the chemical should be used?

src/en/space-station-14/departments/medical/proposals/chemistry-rework.md
```diff
@@ -0,0 +1,61 @@
...
+Base OD of every reagent will be 10u, but it can be randomly decraised during reactions (and should be checked on practice).
+
+### Names of random reagents
+Reagent's name will be generated from random prefixs and suffixs. Like a re, de, fi, gum for prefixs. And lin, line, valine, ac for suffixs.
+
+### Random reagents in other places
+Random reagents also can be obtained in other places: ordered in cargo, gained from artefacts and slimes.
```
Dragonjspider, line 22: So you can get random reagents, what exactly does this mean?
Would there be a list of set reactions at the start of each round, randomly assigned to random Reagents?
(Every round has a reagent that causes another reagent to explode, likely to find the same reagent twice)

Or is every reagent completely randomized with a set of possible reactions?
(Chances of running into the same reagent twice are much lower)

I'm also going to assume here that you're referring to the liquid anomaly when you mention slimes, which poses it's own questions about the liquid anomaly.

Also, what about botany?

src/en/space-station-14/departments/medical/proposals/chemistry-rework.md
Dragonjspider: Also, as a side note, does this new system intend to remove or modify all currently existing chems?
If so it might be worth mentioning how some mechanics might be changed to follow this new system.

* Maints pills 
* Liquid anomalies (Would Reagents like Bicaradine or Omnizine still exist in this case?
* Grinding/Juicing things like Steel and Plasma, and lizard plushies
* Reagents like Lead (How would they interact with the new system, and their ingestion)
* Reagents with specific use cases (Cognizine for example)
* Modular Chemical Grenades (Smoke and Foam)

I think those are all I can think of at the moment...

slarticodefast:
Game design discussion aside - this proposal does not talk about the technical requirements at all. At the moment reagents and reactions are prototypes, meaning they cannot be randomized. So this will require a major refactor. Jez was working on something like that, but someone will have to finish that before anything in this doc could be implemented.

PR #420, opened 2025-03-19
Virology 2.0

AjexRose:
This proposal is for the return of virology and diseases, however it does not include a dedicated role for such
CURRENTLY JUST A DRAFT, ALL THINGS SUBJECT TO CHANGE

### DIFF
--- a/src/en/space-station-14/departments/medical/proposals/virology.md
+++ b/src/en/space-station-14/departments/medical/proposals/virology.md
@@ -1,0 +1,235 @@
+# Virology
+
+| Designers | Implemented | GitHub Links |
+|-----------|---|---|
+| AjexRose  | :x: No | TBD |
+
+
+## Overview
+A replacement/reintegration of virology that hopefully isn’t as bad as before.
+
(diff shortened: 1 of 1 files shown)


walksanatora:
[link to rendered](https://github.com/AjexRose/docs/blob/viro-proposal/src/en/space-station-14/departments/medical/proposals/virology.md)

ArcaneOcto:
This sounds awesome! Although I am curious as to how a virus is introduced to a given round. Do a set number of people get infected round start or as a midround event? Or can a virus emerge naturally? Such as from exposure to stagnant puddles, corpses, alien artifacts, or spoiled food? 

As a side note, I wonder if there's a way to collaborate more with other departments in the future. Perhaps Science can gain points somehow if Virology develops particularly strong or unique virus strains. Or maybe complete vaccines and petri dishes could be sold at Cargo. With this proposal, people in Virology often will end up with their work amounting to nothing as viruses don't show up every round. Though this might not matter at all for an initial design doc.

PR #422, opened 2025-03-20
Station Powered Energy Stimulation System - aka the "Big Laser", "Power Transmission Laser", "SPESS"

Compilatron144:
A proposal for a Power Transmission Laser, rebranded as the SPESS with additional functionality in addition to power transmission.

### DIFF
--- a/src/en/space-station-14/departments/engineering/proposals/SPESS.md
+++ b/src/en/space-station-14/departments/engineering/proposals/SPESS.md
@@ -1,0 +1,155 @@
+# Station Powered Energy Stimulation System - aka the "Big Laser"
+
+| Designers | Implemented | GitHub Links |
+|---|---|---|
+| Compilatron | :x: No | TBD |
+
+## Overview
+
+The "Station Powered Energy Stimulation System" (which will be referred to as the "SPESS") is an engineering machine which serves the function of producing spesos or materials in exchange for power and sufficient cooling.
+
+## Background
+
+As of currently, there are a few ways engineers can boost a station's power production, but there is no reason to do so since excess power is voided. This means that a 400kw singularity and a 6MW TEG are interchangeable as far as powering the station goes.
+
+The solution existed in 13, in the form of the Power Transfer Laser, a machine which would beam energy off into deep space in exchange for money. The purpose of the PTL is to reward engineers for going above and beyond - exchanging extreme amounts of power for money. 
+
+However, the PTL suffers from being a one dimensional machine. Although the source of power may differ, the PTL doesn't offer any interactivity besides pumping more power into it. The SPESS intends to diverge from this by offering it's own mechanics to optimise for.
+
+## Features to be added
+
+### Reflectors
+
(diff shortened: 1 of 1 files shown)


walksanatora:
[link to rendered](https://github.com/Compilatron144/docs/blob/patch-1/src/en/space-station-14/departments/engineering/proposals/SPESS.md)

walksanatora:
my stupid 2 cents: make it instead of a emag interaction (we have enough. and what even is being cryptographically broken here) make it wires to be cut. so you can cut each individual failsafe and the lockout wire. could also open up a stupid idea for engie on nukies. if you see the nukie ship. beam it.

ArcaneOcto:
As a Chief Engineer, I want this so bad. It would give me and my department more things to do after power gen is secured! Much more interesting and rewarding than upgrading solars and making maints cameras imo

ArcaneOcto:
> my stupid 2 cents: make it instead of a emag interaction (we have enough. and what even is being cryptographically broken here) make it wires to be cut. so you can cut each individual failsafe and the lockout wire. could also open up a stupid idea for engie on nukies. if you see the nukie ship. beam it.

I don't see why I can't be both, like the Particle Acceleraror.

walksanatora:
> I don't see why it can't be both, like the Particle Acceleraror.

wait PA can be both emagged and wirecut? huh TIL.

blackknight954:
A friend and myself have started working on getting the PTL working in 14, it'll likely only include the laser itself and not the mirrors or splitter, but it's something at least

will link the PR when it's up.

Errant-4:
I have a big design issue with this. In 13, it might be an acceptable inconvenience to have the beam be present, because there is no actual "traffic" around the station. But here, it's effectively an infinite length wall cutting the space around the station in two, since we only have two dimensions to work with so you can't fly over/under it

(It would also be a bit strange that the laser's target is at a fixed point, but the "local star" quickly circles the station, implying that the station if rotating. Is this also how it works in 13?

blackknight954:
> A less important point, but iIt would also be a bit strange that the laser's target is at a fixed point, but the "local star" quickly circles the station, implying that the station if rotating. Is this also how it works in 13?

Yes, it is

Can look into getting some sort of mirror working as well, if you wish.

walksanatora:
so is this in limbo? being discussed? or is it gonna get closed. engie needs a "lets scale power to absurd degrees" objective and soon.

ArtisticRoomba:
> so is this in limbo? being discussed? or is it gonna get closed. engie needs a "lets scale power to absurd degrees" objective and soon.

starman-superman

Compilatron144:
> so is this in limbo? being discussed? or is it gonna get closed. engie needs a "lets scale power to absurd degrees" objective and soon.

It's engi content, thus highly controversial for adding something called "content"

walksanatora:
it has been a month. and the big laser remains in limbo 😔

blackknight954:
> it has been a month. and the big laser remains in limbo 😔

https://github.com/new-frontiers-14/frontier-station-14/pull/3552#issuecomment-2992617670

Frontier just finished something similllar to it, if someone would like to try and adapt it/port it

Not 100% sure how AGPL and MIT licensed content differ apart from the licensing.

PR #424, opened 2025-03-23
Medcode Writeup

FairlySadPanda:
# Medcode Writeup

(Wait wasn't there a Github template for this??)

A PR to cover some pages that cover "medcode" - specifically Body and the code that interacts with and relates to Body that defines what a spessman is.

I'm opening this as a PR in case folks want to merge it but I'll be adding to this for a bit, would love help and feedback, and need to at the end update references so this is actually linked to inside the docs site.


Partmedia:
Thanks for your docs contribution.

I would suggest:
- Rephrasing things like "Organs are handled extremely inconsistently."
- Using formatting to show when the documentation references the name of a system, e.g. `BodySystem`
- Organizing the high level overview to make it a bit more skimmable

Feel free to improve on these in a subsequent PR. Since the docs repo is very much in the spirit of the old development wiki, I'm going to go ahead and merge this. You may want to consider adding a link to this page.

FairlySadPanda:
Hey notafet,

Was not expecting an insta-merge, ha ha

Yeah this will need a cycle of iteration, I have to do the legwork to outline what everything does first and then go in and explain everything in detail in passes.

PR #435, opened 2025-04-05
add implementation plan requirement to design docs

slarticodefast:
Discussed internally in https://forum.spacestation14.com/t/require-implementation-plan-in-design-doc-proposals/18637
To summarize, a lot of design docs lack critical details about how the technical implementation is planned, and some proposed features were literally impossible in our game engine. Other docs had been accepted a long time ago, but have never been implemented.
Both of these may happen if the author of the doc does not have the required coding knowledge themselves. To solve this we will require authors of new design docs to confirm that they plan on implementing the proposed feature either themselves, or have someone else do it for them.
This helps maintainers to focus on design docs that actually have a chance to be implemented in the foreseeable future.
Existing docs will not be affected.

### DIFF
--- a/src/en/templates/proposal.md
+++ b/src/en/templates/proposal.md
@@ -60,6 +60,6 @@ Consider addressing:
 
 # Technical Considerations
 
+- Does the feature require new systems, UI elements, or refactors of existing ones? Give a short technical outline on how they will be implemented.
 - Are there any anticipated performance impacts?
-- Does the feature require new systems, UI elements, or refactors of existing ones?
 - For required UI elements, give a short description or a mockup of how they should look like (for example a radial menu, actions & alerts, navmaps, or other window types)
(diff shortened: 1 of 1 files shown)


FairlySadPanda:
1. Might be worth providing a slot for which strike team is doing it and trying to get strike teams in general working on topics, not individual people
2. It's a bit weird to have random contributors - who are not part of the project staff - being mentionable as intended implementers of a design doc. This sort of implies that any design doc by non-maintainers has to then be assessed on value judgement of capacity of that person which feels unfair. "We love this design document, but you're a complete chimp of a coder, I guess we can't merge this oh well"

This proposal should be phrased as "design docs need to be tied to a strike team to implement".

slarticodefast:
I think strike teams should stay optional for larger projects.
Also the phrasing might be off putting to authors because they think they need to assemble an entire team just to make a proposal.
`We love this design document, but you're a complete chimp of a coder, I guess we can't merge this oh well`
If it's a good doc someone will be easily willing to work on it. We just want to ensure the docs are actually feasible and not idea guys territory.

PR #441, opened 2025-04-14
update the triage policy

MilonPL:
Makes it clearer when to use which labels, hopefully restrict the use of P1 and P0. 

Vote here: https://forum.spacestation14.com/t/maintainer-policy-change-triage-procedure/19428


Killerqu00:
good stuff

slarticodefast (approved):
Good stuff

ArtisticRoomba (approved):
Makes sense, looks good to me. The priority system was a little bit of a meme before this so I'm glad we're reworking it to be more useful.

FairlySadPanda (changes requested):
P0 needs to also include catastrophic showstoppers outside of what could be considered a "bug". For example, resolving a notice of copyright infringement issued against the project.

P0 is about when you'd expect a Wizard or PM to step in and self-merge.

Errant-4:
Good point. Really, there should be some part in the policy that allows the project to put *any* PR/Issue into a category other than what it would "normally" be in, through some process. Either above or below. The immediately obvious choice is a maintainer vote, say 24h? (With possibly an even simpler "general verbal consensus) shortly after a PR/Issue was opened, just so its easy to change the level of a PR that was initially set to a level maints dont agree with, without running a full vote)

The categories should be the starting point to help initial categorization, not a hard constraint that could get in the way. (Perhaps requiring anything at all to change them is too much, but there should be some consensus behind it?

MilonPL:
The vote has been concluded with 3 approvals and 0 disapprovals. 
Adjusted with a policy allowing two maintainers to change the priority label, based on forums feedback.

PR #442, opened 2025-04-15
New art merge exemption

beck-thompson:
There have been a decent amount of very simple Art PRs that end up sitting around for a long time because they need to maintainer approvals. I feel like they are simple enough that only one approval + art approval is necessary!


EmoGarbage404:
this got incorporated into the new review procedure

PR #456, opened 2025-05-03
feat: graph generation for xeno-arch is expanded and more balanced

Fildrance:
Tried to expand idea of xeno-arch node generation process to add balancing mechanic (so complex triggers, or triggers that have a lot of predessor nodes are going to result in more valuable effects).


iaada (commented):
To be honest I'm pretty confused by the proposal. One word is being used in a lot of different contexts (budget range, virtual budget, actual budget, budget placement). The list of terms helps but I still find it hard to follow.

There's a lot of what feel like repeated descriptions mixed in with new ideas and I couldn't separate the two. It's very implementation specific with a ton of edge cases and just very dense.

If my understanding is correct it goes like this:
- Triggers have a value and a budget range.
- Effects have a value (which can vary based on modifiers).
- For the first node a trigger gets selected. It finds an effect within the trigger's budget range to create the node. That node has an "actual budget" which is the sum of trigger + effect value.
- For the next node a "virtual budget" is created by adding the current depth (times a constant) to the actual value of previous nodes. Then a trigger is chosen within the range of actual and virtual. That trigger has its own budget range which is used to pair an effect.

Fildrance:
Well thats simplified version of thing, but basically yes. And doc have a remark that most of budget stuff is not player-facing. Its techincal details which are not really easy to graps using digrams and text but it would be worse if there were none and there were just plain code with minimal needed xml-doc. 

Part of document explains how effect is scaled/modified (its power changed), which is its own kitchen of a kind, but it is an end goal - make effects feel 'worth' the effort.
Switching from 'virtual budget' (our guess where it should go) and 'actual budget' (what we actually got) is not ideal, i agree, but its a bit hard to describe it without raising a thousand more questions using 'simple' explanation, i feel :(

iaada:
I think it could do with an editing pass by @RemFexxel, but the concept is there and it did come across. It's trying to solve tiered effects so that deeper nodes are more interesting/harder/dangerous. I do think its an important idea to make Artifacts more fun.

My only worry is about microbalance. How are these costs decided? Why should one effect be x budget but another effect by y budget? I understand a harder trigger has a higher budget (and higher budget range?) and more dangerous effects have a higher budget, but more clarity on their costs feels necessary. Part of design docs is to help reviewers be less subjective when looking at new PRs and judging if x budget is appropriate.

As a purely subjective note, I feel like "budget" isn't the right word here. For me the word means "a total to be spent or counted down," but here its only sometimes a total and being used to count up. As a suggestion:

- Budget cost -> intensity
- Budget range -> intensity range
- actual budget -> actual intensity
- virtual budget -> potential intensity


"Intensity" carries the meaning of "danger and power," and a bigger intensity feels more impactful than a bigger budget. By having one of the concepts be that word only, the other concepts become framed in relation to it. "Potential" carries an implication that it's bigger than the total (and in this case it is), whereas "virtual" is a transient word which can mean many things - including in C# where its usage is completely different.

PR #458, opened 2025-05-09
Sentiflower design document

Dead-Wasp:
Document for this https://github.com/space-wizards/space-station-14/pull/33256/files 

This is my first time making a design document, I don't quite understand how to do them

Yeah, I waited six months to do this on Plants vs. Zombies' birthday. What do you know about idée fixe?


Princess-Cheeseballs:
Docs meeting has opted to close this PR for similar reasons to the botany gnomes PR being closed. Botany shouldn't be able to print ghost role workers in isolation and should instead have to source them from other departments, in addition these ghost roles are far easier to produce and far more powerful compared to gnomes.

PR #462, opened 2025-05-18
Pirates Redesigned

RedBookcase:
_"Yarr matey! You are a Pirate, a space-faring loot-grabbing rapscallion! Coordinate with your fellow Scallywags and follow the lead of your dread Captain to raid [Station Name] for plunder! Fulfill your black-market bounties for the dubloons you need to buy better gear, and prepare for the raid of a century!"_

This document outlines a design for **Pirates**, a Major Team-Based Antagonist tasked with repeatedly raiding the station for wealth and glory.

### DIFF
--- a/src/en/space-station-14/round-flow/antagonists/pirates
+++ b/src/en/space-station-14/round-flow/antagonists/pirates
@@ -1,0 +1,61 @@
+# Pirates Proposal
+
+| Designers | Implemented | GitHub Links |
+|---|---|---|
+| thesaladmancer, Doru991, RedBookcase | :x: No | TBD |
+
+## Overview
+
+Pirates are roundstart Team Antagonists tasked with plundering the station to fulfill black-market bounties for dubloons and treasure. Spawning within an isolated Pirate's Bay, they begin with basic equipment and are tasked with raiding the station to steal items to fulfill unique bounties, which provide them with the dubloons they require to buy better and more dangerous gear.
+
+This proposal is intended to repurpose the various Pirate assets within the game to re-implement Pirates as a coherent and interesting Team Antagonist role.
+
+### Gameplay Loop - Preparation
+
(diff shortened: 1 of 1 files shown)


Mixelz:
The possiblity for them to be mid-round spawns is not mentioned in the doc, is the idea that they flat out cannot be mid round spawns? If they are allowed to do so, what sort of compenstation should they be given for joining in late? (Presumably some amount of extra dubloons based on round time and spawning on a dedicated ship instead of the bay).

While the idea of them being hit and run is fun; I do wish there was something in regards to them having their own ship that they can use to fly around the station with, a pirate without their ship doesn't sound like much of a pirate at all. 😔

RedBookcase:
> The possiblity for them to be mid-round spawns is not mentioned in the doc, is the idea that they flat out cannot be mid round spawns? If they are allowed to do so, what sort of compenstation should they be given for joining in late? (Presumably some amount of extra dubloons based on round time and spawning on a dedicated ship instead of the bay).
> 
> While the idea of them being hit and run is fun; I do wish there was something in regards to them having their own ship that they can use to fly around the station with, a pirate without their ship doesn't sound like much of a pirate at all. 😔

We toyed with the idea of having them as midround antags but the main issue is that having a midround antag that relies on multiple players taking up a ghost role is not something we really have atm. If we get the tech for group ghost role raffles we can try it out tho.

Testing them with a ship in admeme rounds revealed that having the pirates be entirely reliant on a ship that would 90% of the time get stolen by salvagers or passengers was a bad idea.

RedBookcase:
> Portals seem a bit out of place for pirates, especially due to the EVA and hardsuits they get
> 
> I'd rather they board the station like actual pirates with jaws of life or AMAGs (either syndie or a new one that only temporarily hacks the door). You could also give them a few explosives but that would be creeping towards nukie territory.

Pirates do not spawn with EVA or hardsuits, only the Captain does.

RedBookcase:
> > Pirates do not spawn with EVA or hardsuits, only the Captain does.
> 
> Can they order deep space EVA and pirate hardsuits?

Later on the plan is to allow them to buy armored hardsuits, but to start with they'll only have an armored overcoat. We don't want Pirates to play like nukies and space everything as a combat tactic, at least not to start with.

PJB3005 (changes requested):
Your design doc is missing an implementation plan following the new requirements: https://github.com/space-wizards/docs/pull/435

PR #466, opened 2025-05-28
Corrections

RemFexxel:
Some of the text looked off to me, so I just went and fixed what I found.


slarticodefast (approved):
Thanks for the cleanup!

PR #473, opened 2025-06-07
Plasma economy

insoPL:
There was some talk about implementing PTL for engies on reddit recently https://wiki.ss13.co/Power_Transmission_Laser
I came up with idea to recreate basic mechanic of turning power into funds but with little bit more inter departamental cooperation and use of already implemented items like plasma.


Cojoke-dot:
I kinda love the idea of getting engi to vgroid so they can frak for plasma… getting it back though sounds like a pain

walksanatora:
personally still in favor of https://github.com/space-wizards/docs/pull/422

insoPL:
> personally still in favor of #422

I like the cooling management and rewarding for steady energy supply part. But at the end of the day it's still press button generate cash and (in that specific proposal)  mats which makes engie self sufficient.

walksanatora:
the SPESS is "pump power in get money/materials out" the more power you put in the better your cooling system has to be. and in the case of ore mining you need your cooling loop to be able to match the target temperature you want for your ores. cause if it slips you are gonna start generating other unwanted materials.

PR #476, opened 2025-06-28
Create bananaonhead.md

FlipBrooke:
A small change that would allow banana peels to be placed on a players head! A real simple addition so I can learn the SS14 workflow.


slarticodefast:
Design docs are only needed for major game changes, like new antagonists, species, departments, jobs or larger reworks of those. Small features like this are totally fine to add without one.
So feel free to just make a pull request to our content repository after you coded this!
https://github.com/space-wizards/space-station-14/pulls

PR #487, opened 2025-08-05
Add Servers On Fire Revert Policy

Simyon264:
wawa


SlamBamActionman:
Minor nitpicking probably but I could this be clarified in relation to reverts considered during the stable review meeting. Should those follow the same steps, or are they using a different method? Just asking since we call those "reverts" too but they may not necessarily be as impactful as how it's described in this policy.

Errant-4:
major nitpick, what is the reasoning for having this be an entire separate policy, instead of being integrated into the hotfix policy, with which it overlaps?

PR #505, opened 2025-08-31
The VVitch

2013HORSEMEATSCANDAL:
Came up with the concept of curses and decided to build a new antag around it because the wizard wouldn't really use it. It's very rough but I'm hoping it's the start of something good.


2013HORSEMEATSCANDAL:
Have yet to come up with a way of breaking a curse, It should be really difficult but not impossible, maybe something with the librarian or chaplain.

PR #507, opened 2025-08-31
Gleep species proposal

Thinbug0:
A proposal for an original fish-snail like alien species. 

visual reference to accompany the doc
showfornow

### DIFF
--- a/src/en/space-station-14/character-species/proposals/Gleep
+++ b/src/en/space-station-14/character-species/proposals/Gleep
@@ -1,0 +1,51 @@
+# Gleep Species
+
+| Designers | Implemented   | GitHub Links                                                              |
+|---|---------------|-----------------------------------------------------------------------------------|
+| Thinbug   | :x: nothin yet|             
+
+## Overview 
+
+This proposal is about the Gleep species as a playable roundstart species. Gleep's are a fish-snail like alien species, easily identified by their shells, which is
+also their center gimmick
+
+(Writing docs is hard so sorry if its a tad messy u.u)
+
+## Background 
+
+Gleeps are an original species, the name is just cute and silly wihout any special meaning
+
+## Why!!
+
+Oki so im giving myself this spot to yap in a very non professional way.
+None of the current species feel the sort of creature vibes i would likey, sure theres dionas which are very inhuman and cool but the vibes they have are very limited and they dont seem very comfy huggable imo.
+So having a squishy soft species with bright colorful eyes making silly alien sounds that can hide in their shell when scared seems very gender comfy so yeah thats why i likey them conceptually. Also why they are
+not a super accurate representation of a snail or fish. I wanna them to feel alien out there weird wacky, monster vibes even, not just a humanoid already known animal
+
+## Design
+
+Going for an inhuman alien vibe, wanting them to feel visually distinct from just a snail or a fish. None of the current species have the sort of smooth soft huggable vibes i would likey (slimes are the closest,
+but are kinda humans 2.0 too human icki), so it would be a different aesthetic to what we have while fitting the vibes of wacky space
+
+Im specially avoiding making them legless, which while it would fit a snail, its something dionas already do and its really visually limiting to remove that clothing slot from the player. Ideally gleeps would be 
+very expressive, rather than being forced into a couple of certain vibes.
+
+## Core Visual Elements
+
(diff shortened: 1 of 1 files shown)


src/en/space-station-14/character-species/proposals/Gleep
lzk228, line 1: change the name from `Gleep` to `gleep.md`

lzk228 (changes requested):

src/en/space-station-14/character-species/proposals/Gleep
Thinbug0: whoops, thanks

ArcaneOcto:
Mechanically I love the idea of a species that can hide in a shell. Would be cool imo if they could fit in bags while in their shell (but take a ton of space). I can imagine a lot of fun to be had when you can have people hide you in bags or throw you around. 
This is also hilarious for RP. Imagine HoP carrying a frightened Gleep captain around during nukeops haha. 

Should there be additional benefits to hiding in your shell? Temperure resist perhaps? 

Would be cool if this species automatically pulls into their shell when crit.

Tbh the only thing I'm not sold on is the name. It's fun and silly, which isn't a bad thing on its own, but it feels a little out of place with the current naming convention for species.

Will definitely keep an eye on this one. There's been so many cool ideas for species now that they're unfrozen!

src/en/space-station-14/character-species/proposals/gleep.md
```diff
@@ -0,0 +1,51 @@
...
+## Core Visual Elements
+
+Sticky boneless smooth soft bodies, sorta fishy vibes while still maintaining a snail-like skin. They may possess gills, bioluminescent spots or different sort of antenna/eyes as markings
+The sounds they make would be generally spacey stereotypical alien sounds (i am having a hard time describing it)
+
+## Species Features
+
+### Shell 
+
+Gleeps have a big shell in their backs, gameplay wise it operates like an unremovable backpack.
```
beck-thompson, line 42: this sounds fun! Maybe it could store stuff but in a unique pattern (a spiral perhaps?)

src/en/space-station-14/character-species/proposals/gleep.md
```diff
@@ -0,0 +1,51 @@
...
+Gleeps have a big shell in their backs, gameplay wise it operates like an unremovable backpack. 
+In return, trough an action they are able to quickly hide inside their shell, losing the ability to move and use items, preventing the usage of radio and greatly increasing their resistances, they can be picked up 
+while in this state.
+
+Exiting their shell would require a lenghtly do after,  during combat it would be a big commitment to enter your shell, as it would be easy for a syndie to keep hitting you, or a secoff to just pick you up and 
+carry you to sec. 
+
+### Waterborn 
+
+Gleeps require to occasionally splash their body with water to keep it hydrated, they will be slown down/take damage if they fail to do so
```
beck-thompson, line 51: I like this idea, but I'm not sure I like that they take damage 🤔

Maybe it could be a buff of some kind? Let me know what you think about something like that

beck-thompson (commented):
I like the sticky aspect of it, can you think of any mechanical implications to that? Maybe if they pick up items without gloves it adds a small amount of slime (glue) to it while they themselves are immune to glue?

ArcaneOcto:
I'm in love with this species concept. Here's a few ideas nobody asked for:

1) I agree that the slime should definitely be a mechanic. It being sticky like glue would be awesome, but what if it was also slightly poisonous? I definitely think you should vomit if you try to ingest gleep slime.

2) Some resistances might be cool, such as a resistance to toxins.

3) Being a snail-esc species I think an interaction with salt is an absolute must. 

4) Hiding in your shell should provide insane defense values, but eject all of the items you had stored. I also like the idea of being an instant action to hide but then being an interuptable do-after to leave.

5) When fully "moistened" it would be funny if they left behind drips of slime on the floor... not much, maybe similar amount to bleeding.

6) I know that some irl creatures (eg: amphibians) require their skin to be moist in order to breath. You could tie that into this species. Gleep could start to suffocate when they dry out, but when moistened they could require significantly less oxygen then other species.

zangothedino:
Hey thought you might like to know that I and a few people in the moffia have been working on 

image
image
image

Thinbug0:
im gonna make tis a draft while i thinkie and like yeah brainblasting thoughtmaxxing

Thinbug0:
my preferences have indeedy changed so ill be doing something elseis

PR #508, opened 2025-09-01
Traitor Corporations [Traitor Objective Rework]

FairlySadPanda:
This is a proposal for a rework of the Traitors game mode. It is not a rework of the Traitor antagonist.

This proposal adjusts some of the current game mechanics surrounding the game mode. It is designed to build a stronger round narrative and give players involved in the round more fun and engaging experiences.

This rework primarily focusses on the role of distinct corporations within the Syndicate and their interactions with NT and each other. It reworks the "every tot for themselves" approach of the current Traitor game mode, allowing for a more diverse set of things that can happen. It also proposes some further work that would build on the intitial rework, that would touch on the Traitor game mode more commonly.

### DIFF
--- a/src/en/space-station-14/round-flow/proposals/traitor-corporations.md
+++ b/src/en/space-station-14/round-flow/proposals/traitor-corporations.md
@@ -1,0 +1,133 @@
+# Syndicate Traitors (Corporations Rework)
+
+| Designers      | Implemented | GitHub Links |
+| -------------- | ----------- | ------------ |
+| FairlySadPanda | :x: No      | TBD          |
+
+# Summary
+
+This is a proposal for a rework of the Traitors game mode. It is not a rework of the Traitor antagonist.
+
+This proposal adjusts some of the current game mechanics surrounding the game mode. It is designed to build a stronger round narrative and give players involved in the round more fun and engaging experiences.
+
+This rework primarily focusses on the role of distinct corporations within the Syndicate and their interactions with NT and each other. It reworks the "every tot for themselves" approach of the current Traitor game mode, allowing for a more diverse set of things that can happen. It also proposes some further work that would build on the intitial rework, that would touch on the Traitor game mode more commonly.
+
+# Why This Isn't A Progressive Objectives Proposal
+
+A "progressive objectives" mechanic for Traitors works like this:
+
+1. The traitor spawns. They are given a simple task to steal some documents, and are given some TC.
+2. The traitor succeeds! They gain TC and get a harder mission to kill a crewmate.
+3. The traitor succeeds! Now they gain more TC and have to kill a Head...
+4. The traitor succeeds! Now they get given even more TC... and when the station runs out of objectives, the traitor gets an objective like "set off the nuke" or "die a glorious death".
+
+The issue with this is that it creates a power ramp for Traitors and pre-assumes a "won" round for Traitors is a destroyed or otherwise massively griefed station. Whilst players do like rounds with maximum death and chaos, Traitor, which tends to be the most common round in rotation, needs to avoid being geared toward excessively chaotic rounds. This tends to put pressure on Security and encourages more of a PVP-style approach to Traitors, when in spirit it is supposed to work a bit more like a social deduction game.
+
+Whilst a progressive objective mechanic might be a fun new alternative to the current Traitor round type, this proposal is focussing on improving the current systems.
+
+# Syndicate Corporations Are Important
+
+The Syndicate is a collection of corporations that are united only by their dislike of Nanotrasen and their distrust of each other. They have continually-shifting internal politics, and even they struggle to keep track of who is even currently signed up as an official Syndicate member.
+
+For the purposes of this document, all Syndicate Corporations are identical, so we'll just be using three common ones as names:
+- Cybersun Technologies
+- Donk Co.
+- Waffle Corp.
+
+Syndicate Corporations are the "true" antagonists of rounds of Traitor, rather than the Traitors themselves. The Traitors work for the Corporations, and the Corporations have objectives that the Traitors need to complete.
+
+## Corporations Have Relations To Other Corporations
+
+Each Syndicate Corporation has one of three relations to each other organization. Each relationship is bi-directional, so both parties have the same relationship with each other.
+
+- Ally: The corporations are currently allied towards a common goal.
+- Neutral: The corporations do not currently share any goals, but have not found reason to cross swords internally.
+- Enemy: The corporations are feuding.
+
+This relationship web is generated randomly each round, and doesn't have any particular weighting. A round could feature a united Syndicate all of whom really want to kick NT's teeth in, or a mafia of competing interests more interested in killing each other.
+
+In addition, the number of Corporations active in a given Traitor round is randomized. A Corporation that is available in the round and can have Traitors assigned to it is called a "candidate".
+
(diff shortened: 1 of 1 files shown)


ilovehans10:
I am a fan of this design. I like how it is based in the reality of the game instead of a gameified system like we have now. 

One thing I worry about is that antags being given the names of other antags allows for plea deals to be made, so even if an antag hasn't done anything yet they can be sold out to Sec.

I also like the inclusion of pink texting, and while I've heard of it conceptually I've not seen how it works when implemented. My one concern is that an antag may think they killed someone and hid the body well enough when really the person was revived and went to centcom in a pod (I know this isn't considered making it to centcom right now, but frankly that's just wrong).

One final thing that I think this would allow is for late join traitors to be implemented more easily. If the number of traitors in a round is chosen based on population, then once the station gets full enough a new hidden traitor slot could open and if someone with traitor enabled joins if their job can be a traitor a coin is rolled and they may be assigned traitor. This would work just like sleeper agents, but would happen more randomly during the round, and would make late joins less metable as safe.

EmoGarbage404 (changes requested):
Attempting a rework of the traitor gamemode without touching the antagonist seems like a generally insane method of operation. The two are obviously and inseparably linked and I don't think a proposal like this can earnestly be considered without giving at least some thought to traitors themselves and how their kits should be reoriented.

I'll outline general concerns section by section:

## Traitors All Have the Same Codewords [...]

There is nothing inherently wrong but I think codewords in general become extremely strange when traitors are just given inherent knowledge of each other. The only purpose they serve is signaling to neutral parties, which seems unwise given the potential of enemies who want to attack you.

As a general note, the "neutral" relation is severely weakened by the looming presence of an "enemy" corporation relationship. Interacting with allies is mutually beneficial, yet exposing yourself to any other traitor doesn't actually benefit you more than just working with your own corporation.

## Syndicate Corporations Can Help Or Hinder Each Other
### Enemies

I'm a longtime hater of traitor v traitor mechanics so allow me outline my position briefly. I believe the purpose of antagonists to be to prompt conflict with the crew via doing various malevolent actions, for the purpose of causing harm unto the crew and giving security things to respond to. Traitors attacking other traitors, while cool, pulls the interaction away from neutral third parties and centralizes a lot of the conflict within the antags themselves. I don't think this is particularly healthy, and would much rather see things that bring more involvement to the crew.

That being said: I also don't feel super compelled by the implementation outlined here. Only presenting a single objective out of the entire set of enemy objectives feels unnecessarily limiting, and really doesn't give the team of traitors enough information to reasonably work against the others. I do also think you would need to require a substantive overhaul of the objectives traitors receive, as objectives which complete themselves (random kills, etc.) would feel excessively unfair if someone cryo'ing in a round resulted in an objective becoming permanently unable to be 100%ed for a team of antagonists.

## Survive and DAGD
I feel like both of these objectives are extremely placeholder and do not offer much to the game. I don't get what they offer to this proposal. Their inclusion just feels like an arbitrary addition in order to retain elements which people would label as "iconic"

## End Of Round
### EOR on MRP
Pinktexting plainly does not make sense with the central objectives of corporations. Does each individual traitor record whether or not they thought they won? How does a traitor accurately deduce if they blocked another corporation from completing their objectives, especially if they are not aware of what objectives they have? I feel like this concept was simply put in here without consideration for how it works with the changes to objectives.

EmoGarbage404:
## SECONDARY CONCERN I FORGOT TO INCLUDE IN MY INITIAL COMMENT

Changing traitors from a primarily-solo antagonist to a primarily team-based one has major balance concerns with regards to the uplink. With the current balance, even if traitors are aware of each other and collude, differences in objectives means that there is at least some incentive to play greedy and keep TC and equipment for yourself. If people are regularly teaming up, have ensured alliances, and  don't have a reasonable fear of betrayal, there will likely be a lot more extreme TC sharing where an individual is able to acquire excess of 20 TC to spend on extremely overpowered nukies-tier gear. I consider this to be a major issue that will very likely skew traitors into being far too aggressive and dominating as a mode.

I think a much more logical way to do it would be to pretty aggressively weaken each individual traitor's gear and abilities in exchange for the increased leverage you get by working with a team of people. If cooperation is the major element you want to emphasize by adding teams to the mode, then cooperation should ideally be the default operation and not provide people with ways of becoming substantially stronger (like it does in the current traitors mode, where solo play is the assumed default).

EthanQix:
While I do love making lore (as it is) more prevalent in SS14, despite what the intro says this proposal does change the Traitor antag significantly, from a primarily solo antag to a primarily team antag. This also makes syndie vs syndie violence more likely, which nudges traitors towards a self fixing problem for the station (which is undesirable). This rework also makes codewords mostly obsolete.

What I'd suggest : keep the corp politics mechanic, but make them more subtle by making the Syndicate keep their agents on a strict need-to-know basis so they have to interact more.
- Traitors shouldn't magically know who their allies are. They should use the existing codewords mechanic to signal and recognize each other. This method encourages social interaction *and* is not foolproof, which is good. Also, it leaves room for the highly funny paperwork mixup situation, where a Traitor has to kill someone from their own corp.
- Traitors shouldn't magically know who their competitors are either, for obvious reasons. However, it could be a fun idea to have them know about *one* codeword of the feuding corps so they can soft-identify competing agents.
- Agents from allied corps should have objectives that encourage cooperation, such as marooning/killing the same target. They don't have to *know* the allied agents' objectives, or even if they are in the same corp or an allied corp.
- Agents from competing corps should have mutually exclusive objectives, such as stealing the same high-value item, or killing someone vs keeping them alive. That someone can be a clueless crew member. Traitors don't have to know they're competing with other traitors.
- Shifting corp alliances midround would be really annoying for Traitors who already completed the corresponding, now obsolete objective.
- The DAGD marker is a *great* idea even if the rest of the proposition doesn't go through.

Mot2332:
I think this rework could make syndicate agents into a proper main game mode, so it COULD be that we have both kinds of syndicate agents, one round they are a sub game mode, working as they are currently implemented, and another round they are the MAIN game mode, and this rework is used

RedBookcase:
I've always liked the idea of making traitor corporations a bigger part of gameplay but I think that making them into full fledged teams removes a lot of the espionage vibe from traitor rounds. I disagree with the idea of corporation members knowing each other from the get-go, but others have noted that concern already.

I think having team-based objectives are interesting but probably more trouble than they're worth for similar reasons, as they allow players to figure out their team size and give them a leg up on coordination. I also think the proposed interaction with corporations and sleeper agents would be pretty annoying - it'd be pretty demoralizing to complete your objectives at the cost of ending up permabrigged, only to find out later that an enemy corporation spawned in a sleeper agent which you were tasked with sabotaging. I think just keeping individual objectives is preferable.

I think there's easier ways to add additional flavor to the different corporations. Corporations could have a list of specialty items in the uplink, one of which is randomly picked per traitor for a "Firesale" with the maximum TC discount (Cybersun providing a discount to Holoparasites, or Donk Co. selling Hyperzine for dirt cheap). Give it a flashy tag in the uplink for a little more color. Each traitor in a corporation would get a different item from the list for their firesale, to prevent something crazy like 4 discounted Juggernaut suits.

walksanatora:
[link to rendered[(https://github.com/FairlySadPanda/honk-docs/blob/hannah/traitor-corporations/src/en/space-station-14/round-flow/proposals/traitor-corporations.md)

PR #509, opened 2025-09-02
Wizard Rework (WIP)

TrixxedHeart:
This rework aims to make the Wizard an antagonist with more direction that aligns with SS14's design philosophy.

It's still highly work in progress and exists to get the ball rolling on discussions to actually make Wizards an interesting antagonist thats fun for all sides in the spirit of the game.

Clear gameplay details have not been written out yet beyond top level details. Hence why it is in a draft state, I want to open a frame of discussion to further develop these details!!!

### DIFF
--- a/src/en/space-station-14/round-flow/proposals/wizard rework.md
+++ b/src/en/space-station-14/round-flow/proposals/wizard rework.md
@@ -1,0 +1,86 @@
+# Wizard Rework
+
+| Designers | Coders | Implemented | GitHub Links |
+|---|---|---|---|
+| TrixxedHeart | TrixxedHeart | :x: No | None |
+
+## Overview
+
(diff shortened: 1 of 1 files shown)


Archisteks:
This can add Spectral Blade, as well as Improved Spectral Blade (and even a particularly devastating magic missile!) for the different wizard levels. 10/10.

EthanQix:
> Introduce Revolution in SS14 without a design doc : mode becomes a bloody mess. Introduce Wizard in SS14 without a design doc : mode becomes a bloody mess. I'm starting to see a pattern here.

I like this concept. Limiting spell choices depending on what wizard archetype is played would give them more identity and clearer objectives, while removing optimized murderboning builds. Also requiring no hardsuit for casting only makes sense - we're playing a wizard here, not some random spessman pleb.

insoPL:
You should consider adding both mana mechanics to limit overall spell spam and rpg-like magic strength score. Then, by giving mage robes that boost his magic strength, you would encourage wearing it while not outright blocking hardsuits. Magic strength should work as a multiplier of spell strength (improving casting time, dmg, mana regen, and range). You could also give higher or lower base magic strength based on pop size, mage type, or luck. Maybe add a way to increase magic strength over time by completing some kind of objective.

PR #517, opened 2025-09-21
Role classification system

chromiumboy:
This proposal outlines the implementation of a role classification system that can be used to better define and clarify expectations around player behavior for antagonist and ghost roles by breaking down broad labels such as 'solo antagonist', 'team antagonist', and 'free agent' into a collection of modular and combinable aspects.

This system can also be used to communicate behavioral expectations to players. Here's a mock-up of a potential UI design:

role-description

### DIFF
--- a/src/en/space-station-14/round-flow/proposals/role-classification-system.md
+++ b/src/en/space-station-14/round-flow/proposals/role-classification-system.md
@@ -10,7 +10,9 @@ Currently, all antagonist and ghost roles are categorized by broad labels, such 
 ## Proposed solution
 Instead of using a small set of predefined labels for describing antagonist and ghost roles, we should consider breaking them down into a set of aspects that can be freely combined, much like the [SCP Anomaly Classification System](https://scp-wiki.wikidot.com/anomaly-classification-system-guide). Having our own role classification system would be very useful from both a design perspective, as well as an aid for players, by providing needed granularity in the descriptions of these roles and as a way of rapidly providing a high-level overview of how different roles are expected to function. 
 
-In this proposal, each antagonist and ghost role is defined by an associated faction (e.g., 'NanoTrasen', 'Syndicate', 'Animal') and four different key aspects: Disruption, Notice, Escalation, and Fraternity. Each aspect has four different ‘levels’ associated with it that modify expectations around player behavior. This system could be readily expanded upon to accommodate crewmember roles as well.
+In this proposal, each antagonist and ghost role is defined by an associated faction (e.g., 'NanoTrasen', 'Syndicate', 'Animal') and four different key aspects: Disruption, Notice, Escalation, and Fraternity. Each aspect has four different ‘levels’ associated with it that modify expectations around player behavior (see below for a more indepth explanation of each of these aspects). This system could be readily expanded upon to accommodate crewmember roles as well.
+
+It should be noted that the proposed system is not intended to replace mechanical enforcement of behavior. For example, Escalation 0 could be mechanically enforced via the pacifism trait. However, being clear with players how a role is intended to function is important for both cultivating the behavior we would like to see from our players, and hopefully reducing attempts to circumvent mechanical enforcement, as well as providing an introduction to the role for newer players and how they can best succeed in the role.
 
 ### Disruption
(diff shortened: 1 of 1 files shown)


src/en/space-station-14/round-flow/proposals/role-classification-system.md
```diff
@@ -0,0 +1,70 @@
...
+
+### Notice
+Notice relates to how noticeable the player's actions should be to others on the station. Note that this aspect may not apply to certain roles, such as silicons.
+
+| Level | Label | Player facing description |
+|-------|-----------|---------------------------|
+| 0 | None | "Being detected would be disastrous for your mission - avoid notice at all costs." |
+| 1 | Low | "You work best when you aren’t attracting much attention. Keep a low profile." |
+| 2 | Moderate | "Being stealthy may be beneficial for achieving your objectives, but you have other tools you can rely on." |
+| 3 | High | "You’re here to here to cause trouble and get everyone’s attention." |
```
IProduceWidgets, line 33: This whole sections feels like a rules bandaid for bad role design. If there are tools available that would result in a playstyle that would be beneficial to the role but detrimental to its design goals then those tools shouldn't be available or their objectives should be realigned to make those actions less attractive.

src/en/space-station-14/round-flow/proposals/role-classification-system.md
```diff
@@ -0,0 +1,70 @@
...
+
+### Fraternity
+Fraternity relates to player cooperation with other faction members.
+
+| Level | Label | Player facing description |
+|-------|-------|---------------------------|
+| 0 | None | "The other members of your faction are your competitors. Either ignore them, or eliminate them." |
+| 1 | Low | "The other members of your faction are your rivals. Temporary alliances might arise if your goals align." |
+| 2 | Moderate | "The other members of your faction are your colleagues. You are welcome to assist them." |
+| 3 | High | "The other members of your faction are your allies. Do whatever you can to support them." |
```
IProduceWidgets, line 53: Implies some method of knowing the faction members. This exists for Zombies, Nukies, and Revs, but it otherwise unclear. With that in mind, I don't know when it would be appropriate for a role to know outside of high.

chromiumboy: There are plenty of ways of discovering who other members of your faction. You mention some, but there others, such as traitor code words and social deduction Some faction members are also going to pretty obvious, like 'Animal' or 'Xenoborg'

src/en/space-station-14/round-flow/proposals/role-classification-system.md
chromiumboy: I agree in that roles should have tools that naturally move them towards the desired playstyles. But it doesn't hurt to remind players of it. For example, ninjas are a rare role that players don't often get a chance to play. With their powerful combat capabilities, new ninja players might be tempted to bite of more than they can chew, and get taken out quickly. Telling players straight up 'hey, try keeping a low profile' nudges them towards how we want to act without them having to learn some things the hard way

chromiumboy:
> I think a lot of these classifications are overly restricting, and that the restrictions they imply are devoid of mechanical backing when really they should be primarily reinforced by gameplay mechanics.

Then it's just a matter of tweaking the descriptions to meet our needs. They are in no way set in stone. Some of these can certainly be supplemented by mechanical reinforcement. For example, combining Escalation 0 with pacifism. But mechanical reinforcements can be circumvented. Being clear to players, 'hey, even if you find a way around the mecahnic reinforcement, here's how we expect you behave' is important for establishing a good player culture from the offset

Errant-4:
1. Note that role subtypes already exist to further clarify what specific role someone is playing. However, right now they are intentionally only displayed to admins, because

2. The "every player is sorted into a small number of possible roles" was a *feature*.  This was intended by the then-headmin to make it clear at all times what is expected of someone, by keeping the possible options limited and well-defined. 


This kinda feels like a page of the Guidebook fell out and landed in the Character window. Should we not rather link the player there?

Mot2332:
If I understand correctly, this information is meant for the **player** themselves, not the **character** they are playing in game. While the ghost role pop ups are OOC information, this is IC information due to being in the character window, which all seems fine, but it is somewhat unclear.

PR #521, opened 2025-09-26
Update docs for arm64

VasilisThePikachu:


slarticodefast:
Maybe mention the RT version number from which on it was supported?

PR #525, opened 2025-10-04
[Proposal] Verb-Centric Interactions

scattagain:
A major refactor interactions to be based around verbs.

Initially was going to be a pr to the main project for a small-ish feature I wanted to see, became a proposal for a project-wide rework of the interaction system.

### DIFF
--- a/src/en/general-proposals/interactionrework.md
+++ b/src/en/general-proposals/interactionrework.md
@@ -1,0 +1,51 @@
+# Verb-Centric Interactions
+
+| Designers | Coders | Implemented | GitHub Links |
+|---|---|---|---|
+| scattagain | scattagain | :x: No | TBD |
+
+## Overview
+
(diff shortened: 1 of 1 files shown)


slarticodefast:
PJB had a similar suggestion before [here](https://forum.spacestation14.com/t/interactions-rework/22758), you might want to link this doc in that thread.

The main problem I see with this is prediction, see https://github.com/space-wizards/space-station-14/issues/39286
A lot of interactions are currently not predicted at all, and their components only exist on the server. So the verb showing you the interaction with them will pop up late, which is extremely annoying to interact with and a major dealbreaker for a verb-focused interaction in my opinion.
472761535-dcf5dbe2-08f9-4774-83b3-11514c774acd

Most of the interactions will just have to be predicted over time, which means a major cleanup effort (see https://github.com/space-wizards/space-station-14/issues/40655), but some stuff like power code or atmos is basically impossible to predict.

So this is a problem I would like to see addressed somehow.

PR #527, opened 2025-10-07
Fix aa.da maintainer listing

ToastEnjoyer:
aa.da is not janet


ToastEnjoyer:
janet is a skinwalker

PR #529, opened 2025-10-10
Clarify endorsement definition in MRP policy

Lyroth001:
Tweaks endorsement definition to match what admins allow endorsing to override.


slarticodefast:
I'll do a quick headmin ping to confirm.

PR #530, opened 2025-10-13
The Blob

ActiveMammmoth:
Basic Blob design document. Saw it didn't have one so I figured it could use one. I'm willing to start coding it if given the go-ahead and would *greatly* appreciate any help given or if a maintainer would like to lead the charge.


Pok27 (commented):
You need to add your proposal in the `SUMMARY.md` file so that it will be visible on the web page

I only have questions about style at this point. It might be better to structure some sections using lists (* or #), for example, as in this document https://github.com/space-wizards/docs/blob/master/src/en/space-station-14/round-flow/antagonists/pursuer.md

psykana:
Hi @ActiveMammmoth,
I reviewed the doc, and it's a great start to bringing this classic to SS14 💪
Are you still willing to work on this?
With maintainer support we might even put together a blob workgroup, but this has to be merged first before any development takes place.

ActiveMammmoth:
@psykana I'd be totally willing to work on it when the current... 'situation'... is resolved definitively.

PR #520, opened 2025-09-26
Diseases Design Document 

Pok27:
Basic disease design document (without mutations or creations)
Needed for space-wizards/space-station-14#40545

PR #543, opened 2025-10-26
Create Sickness System Proposal

JesterX666:
New sickness system proposal

### DIFF
--- a/Sickness System Proposal
+++ b/Sickness System Proposal
@@ -1,0 +1,297 @@
+# Sickness System
+
+| Designers | Coders | Implemented | GitHub Links |
+|---|---|---|---|
+| JesterX | JesterX | No | TBD |
+
+---
+
+## Overview
+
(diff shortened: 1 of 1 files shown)


Princess-Cheeseballs:
Good morning. The medical workgroup has looked over this document and decided to reject it for the following reasons:
- 1. https://github.com/space-wizards/docs/pull/520 a doc for this already exists and is in the final stages of review
- 2. This doc reads more like a list of features to be added rather than an overarching design/architecture that can be built upon
- 3. It doesn't have the detail I would want out of a design doc of this magnitude. Specifically it describes diseases as admin only and vastly underestimates the admin workload the comes with this. (Rules for self-antag for purposefully spreading diseases? Space law implications for viral infection?)
- 4. The Psychological diseases category is arguably offensive in it being lumped in with viral and bacterial infections.

Thank you for your effort but this document is denied.

PR #545, opened 2025-11-01
Rewording and Fixes

RemFexxel:
Some of the docs had spelling mistakes or text that could be reworded more effectively.


slarticodefast (approved):
Looks good to me, thanks for the cleanup!
I'll let another maintainer double check and do the merge.

PR #552, opened 2025-11-10
Dynamic Rooms

FairlySadPanda:
Space stations are both big floating metal structures which you can stand on ("grids") and places of habitation with hallways, maintenance tunnels, departments, offices, and so on.

In Engineering, it's useful to think about a station, be it a static one or a shuttle, as "zones" - an area provided atmosphere that is airtight is one zone, or maybe an area serviced by one substation is one zone. Security and Command can think about the station as being broken up into areas based on access. But every department and every agent on the station can agree that the station is _definitely_ broken up into rooms.

This document discusses how to define rooms, what their impact on parts of the game are, and how they can be modelled to build features on. It'll explore some ideas for features that could use rooms.

### DIFF
--- a/src/en/space-station-14/mapping/proposals/rooms.md
+++ b/src/en/space-station-14/mapping/proposals/rooms.md
@@ -1,0 +1,188 @@
+Space stations are both big floating metal structures which you can stand on ("grids") and places of habitation with hallways, maintenance tunnels, departments, offices, and so on.
+
+In Engineering, it's useful to think about a station, be it a static one or a shuttle, as "zones" - an area provided atmosphere that is airtight is one zone, or maybe an area serviced by one substation is one zone. Security and Command can think about the station as being broken up into areas based on access. But every department and every agent on the station can agree that the station is _definitely_ broken up into rooms.
+
+This document discusses how to define rooms, what their impact on parts of the game are, and how they can be modelled to build features on. It'll explore some ideas for features that could use rooms.
+
+# Foundations
+
+## Definitions
+
+A grid is a set of tiles that together are connected together as the "floor". 
+A wall is a usually-impassable, usually-airtight object that is bolted onto a grid.
+A room is a set of tiles in a grid that are contiguous and surrounded completely by walls.
+
+```
+Key: 
+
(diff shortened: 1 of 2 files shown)


src/en/space-station-14/mapping/proposals/rooms.md
```diff
@@ -0,0 +1,188 @@
...
+
+This means that the room structure needs to be configured in ways that players can also configure. This makes sense. If the players decide to build a really nice shuttle, you'd assume they'd want to define that shuttle's rooms. If a Head of Security decides to revise access to parts of the station, it would not be useful if rooms could be permanently deleted.
+
+As such, we need a premade room model of the station, set at round start. If the station's walls and structure change, rooms should be marked as invalid in the model, and those rooms require maintenance. The game internally keeps a track of what rooms exist on the station, but converting those rooms into the model is first done by the mapper, and then left up to the station's crew to maintain.
+
+This is a job that feels appropriate for the Station AI to keep on top of. Keeping an eye on new rooms bolted onto the hull and adding them to the room model, and fixing the model when the bar is remodelled, are tasks which can't, and shouldn't, be left to an automatic system running on some update loop. It would be always broken in annoying ways.
```
slarticodefast: I'm not sure if the room declaration should be something that can directly be chosed by players, especially if it has an impact on the round - like antag spawns. Instead I would prefer it as a Rimworld/ONI style point system that automatically adjust to the structures inside a room.
Medical machines give points towards a medbay. Lots of closets and catwalks - this is maints. Hydroponic trays - Botany. And so on. The players can still inderectly influence this system, but they will have to entirely refunction the room for it.

Hitlinemoss:
> This is currently not ideal, since the ambient music for maints is often not playing because you almost always have a substation room or a firelock with engineering access close to you - turning all of maints into an engineering area.

We have ambient music for maints?????

The only ambient music I've ever heard in this game is the tracks that play when floating in space. I always thought this was an intentional design decision to improve immersion (e.g, if you hear music on the station then it's most likely from an ingame instrument or object rather than being non-diagetic).

PR #560, opened 2025-12-11
[Proposal] Animation State Machine

Ataman:
This PR contains a proposal for solving the lack of a generalized system for running animations on entities (mostly those using `SpriteComponent`) without having to implement another component/system pair every single time. I'm submitting this as a draft since I'm uncertain how much it should go into the details and was hoping for feedback. This is of course also a chance for everybody to ask questions and add any wishes, requirements or other ideas I might have missed.

[Google Doc with further details](https://docs.google.com/document/d/1mf8Y7cMwGgKG3wiBuEuzWC4UNHURsRVxTTsypZPsVAk/edit?usp=sharing) (Contains details and changes that haven't made it into the proposal yet).

Hopping mice example code: https://github.com/Ataman/space-station-14/blob/asm/Content.Client/Animations/StateMachine/AnimationStateMachineActions/Animations/HoppingAnimationStateMachineAnimationAction.cs

<details>
<summary>Showcase: Waddling clown runs into space.</summary>
In this clip we can see a clown equipped with the state machine using three states:
First, they're standing still and the default state is active (null state in this case). Once the player starts walking, the waddling state becomes active since the IsWalking condition now returns true. Finally, due to some unlucky circumstances, the clown stepped into space resulting in the third state getting active since IsWeightless returns true now, while IsWalking no longer does.

https://github.com/user-attachments/assets/f190a630-f17b-417a-a5a1-d5f1c7bc4161
</details>

<details>
<summary>Showcase: Hopping mice</summary>
Only two states are used in this case. Null and Hopping. A random timer between 0.5 to 2 seconds is used to trigger a Hopping state which has the OneShot property set to true, resulting in a single hop before entering the null state again until the next timer hit.

https://github.com/user-attachments/assets/8627cbca-b55c-407c-895b-040125d7f978
</details>


Ataman:
After testing and playing around on my fork for two weeks; I feel confident enough about this proposal. Removing draft status now.

Also, I'm aware that the name (AnimationStateMachine) sounds atrocious. I welcome *any* suggestions.

Ataman:
Given permission, I could implement this as a System/Component pair inside RobustToolbox itself.

I believe this system could also alleviate some of the headaches about animations by adding some validation code inside the ASM-System itself. (For example: Validating that separate state machines aren't trying to animate the same component simultaneously).

Ataman:
Closing this since the proposal is outdated and probably wasn't necessary in the first place.

PR #565, opened 2025-12-21
Changeling doc rewrite

slarticodefast:
The rewrite for the currently planned implementation, which tries to focus more on keeping the original horror movie theme and stealth and deception based abilities.

This document will be updated along with the changeling's development as some features will need to be playtested to see how they affect the roundflow.

The current development status can be found here: https://github.com/space-wizards/space-station-14/issues/39439


SuperGDPWYL:
I really liked changelings only having to survive, but survival requiring murder in order to not die yourself. It made them pragmatic and opportunistic, and also meant they didn't have the option to be boring. Bit sad to see the objectives replaced with just basic ss13 ling target objectives. I feel like there's far too much overlap with traitor in those.

PR #571, opened 2025-12-26
Updated list conventions

Atakku:
Updated list conventions for better downstream experience

PRs updating parent declarations:
- space-wizards/space-station-14#42071
- space-wizards/space-station-14#42072
- space-wizards/space-station-14#42073
- space-wizards/space-station-14#42074
- space-wizards/space-station-14#42075
- space-wizards/space-station-14#42080
- space-wizards/space-station-14#42081
- space-wizards/space-station-14#42082
- space-wizards/space-station-14#42083
- space-wizards/space-station-14#42084
- space-wizards/space-station-14#42085
- space-wizards/space-station-14#42086
- space-wizards/space-station-14#42087
- space-wizards/space-station-14#42088
- space-wizards/space-station-14#42079

PRs updating category declarations:
- TODO

### DIFF
--- a/src/en/general-development/codebase-info/conventions.md
+++ b/src/en/general-development/codebase-info/conventions.md
@@ -710,7 +710,7 @@ You may use `PhysicsComponent` static body anchoring but *only* if you know what
 ```
 - Don't specify textures in abstract prototypes/parents.
 - You should declare the first prototype block in this order: `type` > `abstract` > `parent` > `id` > `categories` > `name` > `suffix` > `description` > `components.`
-- Use inline lists for categories and regular lists for everything else:
+- Use inline lists for empty overrides and regular lists for everything else:
   ```yaml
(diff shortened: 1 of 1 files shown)


beck-thompson (changes requested):
I assume this is to reduce merge conflicts?

This will cause merge conflicts upstream and cause downstreams a lot of pain initially so you need very solid justification. I'm also not sold on this being better for downstreams. If it does indeed reduce merge conflicts the risk it brings is silent bugs. At least now if a parent changes downstreams will be notified by the merge conflict and can take action if if changes how the item works, but now (if this gets past having merge conflicts) it will be a silent failure. That could be a worthwhile tradeoff, but you need to justify it in in writing. Also be sure to add this same justification in the actual `conventions.md` as well so people realize why its done this way and not just cargo cult follow it.

PR #575, opened 2025-12-31
Shipbreaker Salvage - Salvage Rework Proposal

SlamBamActionman:
🚀 + 🔨 = 💰 

This is a design document for a Salvage rework proposal titled Shipbreaker Salvage, or BreakSalv! Because you break things apart, and you get to take breaks, and maybe there's breakfast at the end. Or it breaks the game! Enjoy!

Thank you to Slartibartfast and Minemoder for reviewing an early version and providing feedback.

PR #577, opened 2026-01-01
Onsite Mining (Tower Defense Mining)

pcaessayrs:
This is a (partial-) salvage rework to turn mining into a fun tower defense game; I also opine that salvaging should be split off and dealt with separately (see What This Design Doesn't Do).

This actually has a lot in common with #575 but was written separately - if @SlamBamActionman wants to combine or source ideas or rationale from this that is okay too (they are both structured as "magnet -> pull grid to station -> do things") 

Here are some mockups of how I imagine this would go -- in these pictures, the miners (and an engineer) have constructed a small labyrinth for the mobs (carps and ore monsters) and defend it using emitters and weapons. The gravgen is a stand-in for the drill.
oh yeah
aahelp

### DIFF
--- a/src/SUMMARY.md
+++ b/src/SUMMARY.md
@@ -229,6 +229,7 @@ Space Station 14
 		- [Proposals]()
 			- [Salvage Proposal](en/space-station-14/departments/cargo/proposals/salvage-proposal.md)
 			- [Mail Deliveries](en/space-station-14/departments/cargo/proposals/mail-delivery.md)
+    	    - [Onsite Mining](en/space-station-14/departments/cargo/proposals/onsite-mining.md)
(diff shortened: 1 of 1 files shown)


CollectionOfBones128:
Personally as a salvager who likes mining I don't think turning it into a tower defence sounds very fun.
Mining has always been the lower PvE greater environmental threat arm of Salvage, and I think it could better benefit from furthering this, like through cave-in mechanics, radioactive clusters, foam pockets, etc.

PJB3005:
After yet another discussion on staff Discord I came to the conclusion Salvage is just never going to work again. 

Foot down time. Salvage is dead.

PR #579, opened 2026-01-04
[Proposal] Minor Antagonist: Scavenger Mice

DigitalThaumaturge:
I've been hearing a lot of complaints about the variety in ghost roles lately, and I've noticed that there tends not to be much antagonistic activity before 15-20 minutes into a round. To solve those problems...

Introducing: Scavengers! Large mice that steal generic resources from the station and serve as a bit of a "warm-up" antagonist.


DigitalThaumaturge:
> What is the intent behind making the evolved scavengers lose their antagonist status and making them into free agents ?

Once evolved, they don't really have any objectives. Evolving is their win condition, so everything afterward is just one big victory lap.

And at that point, getting chummy with the crew should be just as difficult as antagonising them, given that you're still a loot piñata full of stolen goods.

PR #580, opened 2026-01-06
Add prediction guide

slarticodefast:
Adds a guide to prediction, along with a few code examples and gifs showcasing the results.
I tried writing down everything that needs to be known, let me know if anything is wrong or unclear.

Draft until I had a few contributors and maintainers proofreading this and give feedback.
Especially the section on ApplyingState seems like it could be explained a little better, but it's a complicated topic.

ToDo in a future PR: Update the `Basic Networking and you` doc to include sections on field deltas, `NetSerializableAttribute` and how `EntityUids` are converted into `NetEntities` by the source generator.

### DIFF
--- a/src/en/ss14-by-example/prediction-guide.md
+++ b/src/en/ss14-by-example/prediction-guide.md
@@ -52,7 +52,7 @@ When setting breakpoints or printing to the console you will notice that the pre
 ## Prediction Code example
 Let's look at a simple unpredicted example component and predict it.
 
-```C#
+```csharp
 // In Content.Server/PredictionExample/PredictionExampleComponent.cs
 
 using Robust.Shared.Audio;
@@ -81,7 +81,7 @@ public sealed partial class PredictionExampleComponent : Component
 }
 ```
 
-```C#
+```csharp
 // In Content.Server/PredictionExample/PredictionExampleSystem.cs
 
 using Content.Shared.Examine;
@@ -165,7 +165,7 @@ If we look at this in-game we notice the verb, examination text, popup and sound
 
 Now let's move the code to shared and network and predict it.
 
-```C#
+```csharp
 // In Content.Shared/PredictionExample/PredictionExampleComponent.cs
 
 using Robust.Shared.Audio;
@@ -194,7 +194,7 @@ public sealed partial class PredictionExampleComponent : Component
 }
 
 ```
-```C#
+```csharp
 // In Content.Shared/PredictionExample/PredictionExampleSystem.cs
 
 using Content.Shared.Examine;
@@ -274,7 +274,7 @@ The result is much more responsive without the delay for the popup, audio or UI 
 Shared code can only call other shared code, whereas server-side code can use both server-side and shared code. This means if you want to predict an EntitySystem you will need predict all its dependencies first so that you can use them in Shared, which often turns prediction PRs into much larger tasks than initially expected.
 
 Some systems cannot be predicted, but you might still want to call some API methods that are only available on the server from Shared. To work around this you can add an empty virtual API method in the corresponding shared system and override it on the server. Here's an example from [`SharedExplosionSystem`](https://github.com/space-wizards/space-station-14/blob/master/Content.Shared/Explosion/EntitySystems/SharedExplosionSystem.cs):
-```C#
+```csharp
 // In Content.Shared/Explosion/EntitySystems/SharedExplosionSystem.cs
 // This method is empty and does nothing on the client.
 public virtual void TriggerExplosive(EntityUid uid, ExplosiveComponent? explosive = null, bool delete = true, float? totalIntensity = null, float? radius = null, EntityUid? user = null)
@@ -341,7 +341,7 @@ B) The insertion of the entity into the container is not predicted (for example 
 This is done that way to allow a client to update UIs even if they did not predict the event, for example the storage window of your backpack, your hand indicator or the damage overlay, but it also may cause problems during subscriptions as any changes done inside them are already networked separately within the same game state, meaning they will be applied multiple times, causing mispredicts.
 
 Let's look at an example from `GlueSystem`
-```C#
+```csharp
 // GluedComponent will make an item temporarily unremovable if you pick it up.
 private void OnHandPickUp(Entity<GluedComponent> entity, ref GotEquippedHandEvent args)
 {
@@ -385,7 +385,7 @@ The most common ones are
 
 ## Predicted update loop example
 A lot of old code is accumulating frametime inside update loops to decide when to next run it.
-```C#
+```csharp
 /// <summary>
 /// Unpredicted example.
 /// </summary>
@@ -430,7 +430,7 @@ public sealed class UpdateLoopExampleSystem : EntitySystem
 ```
 This is bad for prediction, as dirtying the accumulator datafield every single tick would be expensive due to the network load if we repeatedly have to send game state updates from the server. So instead of accumulating frame time we use a time stamp indicating when the next update is supposed to happen and compare that with the current server time.
 
-```C#
+```csharp
 /// <summary>
 /// Predicted example.
 /// We need to network the component for this.
@@ -547,7 +547,7 @@ Here is an example of a mispredict happening when gibbing someone, so that you k
 In the future Robust Toolbox will have methods for predicted randomness, but at the time of writing the [PR for RandomPredicted](https://github.com/space-wizards/RobustToolbox/pull/5849) has not been merged yet.
 As a workaround you can use a new `System.Random` instance and set the seed to something the server and client agree on, for example a combination of an entity's `NetEntity` id and the current game tick (if you would only the game tick here then all randomness within the same game tick would yield the same result, so we need both).
 
-```C#
+```csharp
 // EntitySystem dependencies:
 // [Dependency] private readonly IGameTiming _timing = default!;
 // [Dependency] private readonly IRobustRandom _random = default!;
@@ -601,7 +601,7 @@ Dirtying and networking is expensive. Avoid dirtying entities every single tick 
 
 When setting a datafield it is recommended to add a guard statement to check if the datafield has a new value before calling `Dirty` so that we only network it when actually necessary.
 
-```C#
+```csharp
 // inside an EntitySystem
 
 public void SetExampleDataField(Entity<ExampleComponent?> ent, int newValue)
@@ -632,7 +632,7 @@ Prediction repeatedly resets any dirtied datafield back to a previous game state
 
 #### Conventions for shared systems and components
 EntitySystems should either be non-abstract and shared, for example:
-```C#
+```csharp
 // In Content.Shared/SomeNamespace/SomeSystem.cs
 public sealed class SomeSystem : EntitySystem
 {
@@ -642,7 +642,7 @@ public sealed class SomeSystem : EntitySystem
 ```
 or have a shared abstract system that is inherited on the server and client.
 
-```C#
+```csharp
 // In Content.Shared/SomeNamespace/SharedSomeSystem.cs
 public abstract class SharedSomeSystem : EntitySystem
 {
@@ -670,7 +670,7 @@ There is currently some weird behaviour with solution entities and PVS, which ma
 
 This debug assert happens if a predicted update loops calls `SharedSolutionContainerSystem.ResolveSolution`, and the entity containing the solution entity leaves PVS range. This should not be happening since entities are paused when moved outside PVS range, meaning the update loop should no longer run on the client, but for some reason this still causes the debug assert and needs a workaround until this is fixed properly in a way that does not need this boilerplate code.
 
-```C#
+```csharp
 private void OnEntRemoved(Entity<UdderComponent> entity, ref EntRemovedFromContainerMessage args)
 {
     // Make sure the removed entity was our contained solution


Pok27:
Very cool job!

src/en/ss14-by-example/prediction-guide.md
```diff
@@ -0,0 +1,687 @@
...
+// This contains the actual code and only runs on the server, but can still be called from shared, since shared code runs on both the server and client.
+public override void TriggerExplosive(EntityUid uid, ExplosiveComponent? explosive = null, bool delete = true, float? totalIntensity = null, float? radius = null, EntityUid? user = null)
+    {
+        // Some code in here that creates the explosion
+    }
+```
+
+## PopupPredicted & PlayPredicted
+
+Using `SharedPopupSystem.PopupEntity` in predicted code will cause the popup to be shown multiple times during prediction. Instead you have to use `PopupPredicted`, which will predict the popup once for the client passed in via the `user` parameter and the server will network the popup to everyone else, excluding that client so that they won't show the popup twice.
```
VerinSenpai: Minor wording nitpick, this should probably say "so that they won't see the popup twice"

VerinSenpai (changes requested):
Partial review. Just some minor wording and grammar nitpicks. Its beautiful 🥺

src/en/ss14-by-example/prediction-guide.md
```diff
@@ -0,0 +1,687 @@
...
+
+## How does prediction work?
+
+Without prediction any input from the client (like keyboard presses or mouse clicks) will be networked to the server, which then simulates the game according to all players' inputs and networks the resulting game state back to each client. The client will only see the results with a noticable delay depending on their ping, which makes the game feel unresponsive and will cause a visual delay for UI elements.
+
+With prediction each client runs its own game simulation of the game according to the local player's inputs and they will immediately see the results without having to wait for the server, hiding any latency. The server holds the authoritative game state, which is what all clients consider to be the "truth" in case they disagree on something.
+
+During prediction the client will constantly time travel, reverting the game state repeatedly to the last known authoritative state sent by the server, and then reapplies the player's inputs, resimulating the game until the next server state comes in. This usually happens about 12 times per game tick.
```
SlamBamActionman: "Time travel" is a correct description but it kinda comes in out of nowhere, and for a beginner may seem out of left field.

I think the entire line should be reworded. It explains the process but the process isn't intuitive enough on its own. What does it mean that the client is repeatedly reverting and resimulating, and why does it do it?

src/en/ss14-by-example/prediction-guide.md
```diff
@@ -0,0 +1,687 @@
...
+
+## Prediction Checklist
+Most of the steps explained above are automatically handled by the game engine for any systems in `Content.Shared`, but you will have to make sure to network your code correctly.
+Follow these steps to make an existing unpredicted `EntitySystem` predicted:
+- Move the relevant components and systems and from `Content.Server` to `Content.Shared`.
+- Add the `NetworkedComponent` and `AutoGenerateComponentState` attributes to the components, and the `AutoNetworkedField` attribute to the datafields that the client needs to know about and that may change after spawning (if they never change from the value in the prototype there is no need to network them) or use manual component states if needed.
+- `EntitySystem`s should either be a single non-abstract system in `Content.Shared`, or an abstract shared system with both the server and client inheriting from it. Even if the client system is empty, make sure it exists or the code won't run on the client and will not be predicted.
+- To be able to make some code predicted, all its dependencies need to be made predicted first, since shared code cannot call server side code in a predicted way.
```
SlamBamActionman:
```suggestion
- To be able to make some code predicted, all its dependencies need to be made predicted first, since shared code cannot call server-side code in a predicted way.
```

src/en/ss14-by-example/prediction-guide.md
```diff
@@ -0,0 +1,687 @@
...
+
+// In Content.Client/SomeNamespace/SomeSystem.cs
+public sealed class SomeSystem : SharedSomeSystem
+{
+    // ...
+}
+```
+Don't forget the client-side system, even if it's empty. Otherwise the client won't be able to instanciate the system class and it will remain unpredicted.
+
+Avoid shared abstract components. Some ancient code is still doing this, but nowadays we just instead put the entire component into `Content.Shared`, even if some datafields are only used on the server or client. This is minimally worse for performance, but makes the code much more readable, simplifies any API methods and makes it easier to use `TryComp` and `Resolve`.
```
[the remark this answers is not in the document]
SlamBamActionman: Wait is serverOnly even mentioned? I think the only thing related to server-only datafields is when you imply not adding AutoNetworkedField to fields that shouldn't be networked.

src/en/ss14-by-example/prediction-guide.md
```diff
@@ -0,0 +1,687 @@
...
+
+See [this guide](basic-networking-and-you.md) for basic information on how to network components, which will be required knowledge for this tutorial. 
+
+## How does prediction work?
+
+Without prediction any input from the client (like keyboard presses or mouse clicks) will be networked to the server, which then simulates the game according to all players' inputs and networks the resulting game state back to each client. The client will only see the results with a noticable delay depending on their ping, which makes the game feel unresponsive and will cause a visual delay for UI elements.
+
+With prediction each client runs its own game simulation of the game according to the local player's inputs and they will immediately see the results without having to wait for the server, hiding any latency. The server holds the authoritative game state, which is what all clients consider to be the "truth" in case they disagree on something.
```
Tayrtahn:
```suggestion
With prediction each client runs its own simulation of the game according to the local player's inputs and they will immediately see the results without having to wait for the server, hiding any latency. The server holds the authoritative game state, which is what all clients consider to be the "truth" in case they disagree on something.
```

src/en/ss14-by-example/prediction-guide.md
```diff
@@ -0,0 +1,687 @@
...
+- `EntitySystem`s should either be a single non-abstract system in `Content.Shared`, or an abstract shared system with both the server and client inheriting from it. Even if the client system is empty, make sure it exists or the code won't run on the client and will not be predicted.
+- To be able to make some code predicted, all its dependencies need to be made predicted first, since shared code cannot call server side code in a predicted way.
+- Dirty the component every time its datafields are changed. This will tell the server to send all networked datafields in that component to the client so they get synced, and it will tell the client to reset that datafield to an earlier value during prediction.
+- For larger components with lots of networked datafields you should be using field deltas (use `DirtyField` for that) to reduce the network load. This will only dirty that datafield instead of the entire component.
```
Tayrtahn: It's also good for components with fields that change at very different rates, like if a component really needs to send one field every second, but the other fields are mostly static.

src/en/ss14-by-example/prediction-guide.md
```diff
@@ -0,0 +1,687 @@
...
+This will need further engine changes in the future, see [Entity spawn prediction v2](https://github.com/space-wizards/RobustToolbox/issues/5845).
+
+#### Deleting
+If you try to use `DeleteEntity` or `QueueDeleteEntity` to let the client delete a non-client-side entity it will cause an error:
+`[ERRO] root: Predicting the deletion of a networked entity`.
+
+If you want to predict entity deletion use the equivalent `PredictedDeleteEntity` or `PredictedQueueDeleteEntity` instead. These will let the client move the entity to nullspace first, which makes it seem to disappear. The server deletes it normally and then networks the deletion to the client.
+
+## IGameTiming.IsFirstTimePredicted
+This returns `true` the first time the code runs , and false in all the following prediction ticks while the client is waiting on the server. On the server this always returns `true` (the server does not predict anything, so it only needs to run the code once anyway). This is usually used internally in the form of a guard statement inside API methods for popups, audio and for some UI code to prevent it from running multiple times or flickering during prediction.
```
Tayrtahn:
```suggestion
This returns `true` the first time the code runs, and false in all the following prediction ticks. On the server this always returns `true` (the server does not predict anything, so it only needs to run the code once anyway). This is usually used internally in the form of a guard statement inside API methods for popups, audio and for some UI code to prevent it from running multiple times or flickering during prediction.
```
I wouldn't say the client is waiting for the server during prediction ticks. Even if the next update has arrived, I think they still get run for each update on the queue (but didn't quote me on that).

src/en/ss14-by-example/prediction-guide.md
```diff
@@ -0,0 +1,687 @@
...
+If your component has multiple networked datafields consider using field deltas to prevent the update loop from networking all of them with each update.
+
+## PVS
+To reduce the networking load on the server it will only network any entities within a certain range (by default a 25x25 square around the client's attached entity). If an entity leaves PVS range then it will be paused so that update loops will no longer run on the client and detached into nullspace on the client until it reenters PVS range.
+
+If you want to see this in action you can fly around as an aghost while having the view zoomed out.
+
+![prediction-guide-pvs.gif](../assets/images/ss14-by-example/prediction-guide-pvs.gif)
+
+Since the server holds the authoritative game state it always has full knowledge about all entities, but the client will be restricted on what information it has available, which has implications for prediction. The client won't be able to predict anything happening outside PVS range, and some code like atmos or power cannot be predicted as a result.
```
Tayrtahn:
```suggestion
Since the server holds the authoritative game state, it always has full knowledge about all entities, but the client will be restricted on what information it has available, which has implications for prediction. The client won't be able to predict anything happening outside of PVS range, and some code like atmos or power cannot be predicted as a result.
```

src/en/ss14-by-example/prediction-guide.md
```diff
@@ -0,0 +1,687 @@
...
+
+#### SessionSpecific
+All components have the `SessionSpecific` bool, which will cause a `ComponentGetStateAttemptEvent` to be raised on the owning entity for each player the component is being networked to and allows you to cancel the networking in a subscription. Note that this comes with some performance overhead since it may raise a lot of events.
+An example for this is [`SharedRevolutionarySystem`](https://github.com/space-wizards/space-station-14/blob/master/Content.Shared/Revolutionary/SharedRevolutionarySystem.cs) which uses this to let only revolutionary players and admins know who else is a revolutionary is. However, the API for this is currently quite annoying to use and requires a lot of boilerplate code. Maybe in the future this could be simplified into a component whitelist to decide who can see a certain session specific component.
```
Tayrtahn:
```suggestion
An example for this is [`SharedRevolutionarySystem`](https://github.com/space-wizards/space-station-14/blob/master/Content.Shared/Revolutionary/SharedRevolutionarySystem.cs) which uses this to let only revolutionary players and admins know who else is a revolutionary. However, the API for this is currently quite annoying to use and requires a lot of boilerplate code. Maybe in the future this could be simplified into a component whitelist to decide who can see a certain session specific component. 
```

src/en/ss14-by-example/prediction-guide.md
```diff
@@ -0,0 +1,687 @@
...
+
+To make sure the BUI is also updated during prediction and not only when receiving the server state you will also need an empty virtual `UpdateUi` method in your shared code that will call the BUI's `Update` method in a client-side override.
+
+A good code example for predicting a BUI using component states can be found in [this PR](https://github.com/space-wizards/space-station-14/pull/33835).
+
+## Things to watch out for
+#### Performance tips
+
+Dirtying and networking is expensive. Avoid dirtying entities every single tick from an update loop. Think about how to minimize the amount of data you have to send. For example in systems like [hunger](https://github.com/space-wizards/space-station-14/blob/master/Content.Shared/Nutrition/EntitySystems/HungerSystem.cs) or [battery charge](https://github.com/space-wizards/space-station-14/blob/master/Content.Shared/Power/EntitySystems/SharedBatterySystem.API.cs) do not network the new hunger or charge value repeatedly, but only send a value at a certain time stamp along with the current rate of change. That way the client can always infer the current value without requiring a new game state to be send.
```
Tayrtahn:
```suggestion
Dirtying and networking is expensive. Avoid dirtying entities every single tick from an update loop. Think about how to minimize the amount of data you have to send. For example in systems like [hunger](https://github.com/space-wizards/space-station-14/blob/master/Content.Shared/Nutrition/EntitySystems/HungerSystem.cs) or [battery charge](https://github.com/space-wizards/space-station-14/blob/master/Content.Shared/Power/EntitySystems/SharedBatterySystem.API.cs) do not network the new hunger or charge value repeatedly, but only send a value at a certain time stamp along with the current rate of change. That way the client can always infer the current value without requiring a new game state to be sent.
```

src/en/ss14-by-example/prediction-guide.md
```diff
@@ -0,0 +1,687 @@
...
+
+#### IRobustCloneable
+
+Prediction repeatedly resets any dirtied datafield back to a previous game state and resimulates the player input while waiting for the server state to come in. However, if your datafield is a [reference type](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/reference-types) it will only reset the reference to that datafield, not datafield's current state. This can lead to [mispredicts](https://github.com/space-wizards/space-station-14/pull/34838) and even [major graphical glitches](https://github.com/space-wizards/space-station-14/issues/38512) if not handled correctly. To fix this make sure to implement the `IRobustCloneable` interface when using autonetworking with any custom reference type you are networking and the source generator will make sure to create a deep copy of the datafield for each game state. A code example for this is the [`Solution`](https://github.com/space-wizards/space-station-14/blob/master/Content.Shared/Chemistry/Components/Solution.cs) class used in the [`SolutionComponent`](https://github.com/space-wizards/space-station-14/blob/master/Content.Shared/Chemistry/Components/SolutionComponent.cs).
```
Tayrtahn:
```suggestion
Prediction repeatedly resets any dirtied datafield back to a previous game state and resimulates the player input. However, if your datafield is a [reference type](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/reference-types) it will only reset the reference to that datafield, not the datafield's current value. This can lead to [mispredicts](https://github.com/space-wizards/space-station-14/pull/34838) and even [major graphical glitches](https://github.com/space-wizards/space-station-14/issues/38512) if not handled correctly. To fix this make sure to implement the `IRobustCloneable` interface when using autonetworking with any custom reference type you are networking and the source generator will make sure to create a deep copy of the datafield for each game state. A code example for this is the [`Solution`](https://github.com/space-wizards/space-station-14/blob/master/Content.Shared/Chemistry/Components/Solution.cs) class used in the [`SolutionComponent`](https://github.com/space-wizards/space-station-14/blob/master/Content.Shared/Chemistry/Components/SolutionComponent.cs).
```

src/en/ss14-by-example/prediction-guide.md
```diff
@@ -0,0 +1,687 @@
...
+    // ...
+}
+
+// In Content.Client/SomeNamespace/SomeSystem.cs
+public sealed class SomeSystem : SharedSomeSystem
+{
+    // ...
+}
+```
+Don't forget the client-side system, even if it's empty. Otherwise the client won't be able to instanciate the system class and it will remain unpredicted.
```
Tayrtahn:
```suggestion
Don't forget the client-side system, even if it's empty. Otherwise the client won't be able to instantiate the system and it will remain unpredicted.
```

Tayrtahn (commented):
Looking great! Thanks for writing all of this up!

src/en/ss14-by-example/prediction-guide.md
```diff
@@ -0,0 +1,687 @@
...
+
+`IsFirstTimePredicted` is currently [misused all across the codebase](https://github.com/space-wizards/space-station-14/issues/41116) to hide mispredicts, so don't trust existing code on this. This does not fix the underlying problem but rather disables prediction.
+
+## IGameTiming.ApplyingState
+This returns `true` while the client either applies a game state networked from the server or rerolls its own game state to a previous one during prediction. This is commonly used in the form of a guard statement to allow server states to be applied properly and prevent code from running when it should not.
+
+To understand why this is needed you have to know that some events are networked along with a components state and are always raised on both the server and client even if not predicted and they are only using `RaiseLocalEvent` and `SubscribeLocalEvent`.
```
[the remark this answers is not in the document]
slarticodefast: I think the "are networked along with a components state" already explains this part, and I don't want to duplicate the part below.

src/en/ss14-by-example/prediction-guide.md
```diff
@@ -0,0 +1,693 @@
...
+```
+Can't resolve "Robust.Shared.GameObjects.MetaDataComponent" on entity 957132/n0D!
+   at System.Environment.get_StackTrace()
+   at Robust.Shared.GameObjects.EntityManager.GetNetEntity(EntityUid uid, MetaDataComponent metadata) in /home/runner/work/space-station-14/space-station-14/RobustToolbox/Robust.Shared/GameObjects/EntityManager.Network.cs:line 186
+...
+```
+
+To prevent this from happening you have to set the datafield back to `null` if the entity is deleted somehow, however this is tricky to keep track of, requiring marker components and a lot of boilerplate code.
+
+At the time of writing the WizDen servers get more than 20000 of these errors each day and resolving them will require some engine changes introducing a `WeakEntityReference`, which references an entity that may or may not be deleted. Once the engine PR is merged the content repository will require some cleanup to get rid of these errors.
```
ArtisticRoomba: Emogarbage closed the WeakEntRef PR as it introduced bad programming practices. I'm probably paraphrasing horribly but the main concern was that WeakEntityRef would be used in a way that would cause bad error handling behavior and system design, thus it's beneficial to instead design systems to deal with entities that may not exist anymore (to an extent).

src/en/ss14-by-example/prediction-guide.md
```diff
@@ -0,0 +1,693 @@
...
+}
+```
+
+This makes sure that the cached reference to the solution entity is reset to null as is leaves PVS range and is detached into nullspace.
+
+See [this issue](https://github.com/space-wizards/space-station-14/issues/42218) for the current state of this problem.
+
+#### Only use NetworkedComponentAttribute for shared components
+Adding `[NetworkedComponent]` to a purely server- or client-side component (i.e. not in `Content.Shared`) does not make any sense since those cannot be networked in the first place.
+However at the time of writing RobustToolbox is not preventing you from doing so and instead of creating a warning or error it will just break silently. The sympoms include random other components no longer being networked to the client, which will cause a huge range of bugs, for example mispredicts and UIs not being populated. So this is something you should keep an eye on during reviews or when writing new code.
```
ArtisticRoomba: Symptoms*

Also Jesus Christ.

ArtisticRoomba (changes requested):
T'was a very good read before bed.

Issue #566, opened 2025-12-22
imgur is used for some pages


deltanedas:
it shouldnt

https://github.com/space-wizards/docs/blob/master/src%2Fen%2Fgeneral-development%2Fsetup%2Fgit-for-the-ss14-developer.md

example page

SphereTenet:
This also means British people just cannot see the docs without a VPN 

<img width="1994" height="1513" alt="Image" src="https://github.com/user-attachments/assets/667f76ff-2520-4ff6-b14b-d46729770673" />

PR #582, opened 2026-01-11
Add workflow for checking for externally hosted images

slarticodefast:
See #566

Simple python script that uses regex to parse all image links and checks if they are relative paths rather than external urls.
There is probably an easier way to do this (maybe it's possible to disable it for the mdbook?) but this works.

Needs the repository cleaned up before merging.

### DIFF
--- a/.github/workflows/test-image-links.yml
+++ b/.github/workflows/test-image-links.yml
@@ -10,5 +10,5 @@ jobs:
     steps:
     - name: Checkout repository
       uses: actions/checkout@v4
-    - name: Run imgage link check
+    - name: Run image link check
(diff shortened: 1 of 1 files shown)


slarticodefast:
@RemFexxel Here is a list of links that need to be fixed in case you want to work on this.
See the failing test log below.

RemFexxel (commented):
Minor spelling mistake.

ArtisticRoomba:
`hedgedoc` should probably be allowed as we host that ourselves.

ArtisticRoomba (approved):
thanks a bunch for the changes. be sure to have a restful break

PR #583, opened 2026-01-12
Remove Externally Hosted Images

RemFexxel:
This pull request addresses https://github.com/space-wizards/docs/issues/566.

Over 50 files contain Imgur links or other externally hosted images. This pull request removes them and replaces each with a self-hosted file.

This may take some time to complete, so this pull request will remain a draft until all links are replaced. If there are any mistakes made, please comment on them.


ArtisticRoomba (changes requested):
Some minor things I saw

In general I'd say it's fine to not remove images hosted on `hedgedoc` because that is something we host.

PR #587, opened 2026-01-20
Telesalvage Proposal

SolventMercury:
I'm not sure what the current state of discussion is on salvage, but it's clear that it's fucked and anything similar to it as it is currently would be Bad. This is my proposal for a salvage-esque role that meshes better with the game as it exists. The gist of it is basically salvage via telescience, salvagers getting a portal opened to a dungeon instead of flying out to something in space.  Would love feedback!

### DIFF
--- a/src/SUMMARY.md
+++ b/src/SUMMARY.md
@@ -226,6 +226,7 @@ Space Station 14
 		- [PR Guidelines]()
 
 		- [Proposals]()
+			- [Telesalvage Proposal](en/space-station-14/departments/cargo/proposals/telesalvage-proposal.md)
 			- [Salvage Proposal](en/space-station-14/departments/cargo/proposals/salvage-proposal.md)
 			- [Mail Deliveries](en/space-station-14/departments/cargo/proposals/mail-delivery.md)
 
--- a/src/en/space-station-14/departments/cargo/proposals/telesalvage-proposal.md
+++ b/src/en/space-station-14/departments/cargo/proposals/telesalvage-proposal.md
@@ -1,0 +1,38 @@
+# Telesalvage
+
+# Salvage
+
+| Designers      | Implemented | GitHub Links |
+|----------------|---|---|
+| SolventMercury | :x: No | TBD |
+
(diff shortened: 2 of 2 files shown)


YoungThugSS14:
IMG_5953
Good doc though

ArtisticRoomba (commented):
from a short read this very much feels like a "draw the rest of the horse" design in some areas. it's good to put salvies on-station and get general players to interact with this content (a big problem of the previous 4 iterations) but you're still stuck with the problem of "how are you going to get the janitor to interact with rimworld".

the document also identifies salvage's off-station problem as contributing to the lack of resources problem, however this design document doesn't even dive into how that problem is solved with this proposal:

> Salvagers are incentivized to spend as little time on-station as possible, especially because of how long it takes to set up a shuttle and start playing the role. This leads to them often failing to distribute resources; conversely, cargo often buys resources like steel even if salvage has already procured large amounts of it.

with the only design point i could find that attempts to somewhat address this being:
> Many items should be useful to departments unrelated to salvage, or useless outside of RP/fun value, like rare wines, undiscovered alien plant seeds, wearable jewelery, or particularly cushy furniture. These items work as an implicit encouragement for salvage to either return to the station and show off their loot, or invite the crew down to admire the haul.

jewelry and other department specific equipment like RTGs exists, and in both of the iterations those rewards existed in, it didn't really motivate salvagers to show it off to the station. *sometimes* you'd get it, sometimes you didn't, more often though it was either ignored and sold. this doesn't strike as a good solution to me

---

in general salvage is pretty antithetical to the game's core design principle of player interaction and i don't really see any iteration where the crew can meaningfully interact with salvage and their mechanics in a way that benefits the station as a whole without doing something horrible to station roundflow at the same time.

i'm starting to push SS14 towards system-based design where in-depth mechanics are available to everyone in some degree, with departments having jobs and gameloops that revolve around those mechanics (see chemistry refactor). i don't really see salvage accomplishing this in any iteration that I see.

the department was fun but i'm ready to see it on its way out.

PR #593, opened 2026-01-29
Bitrunners. DeTGified.

kosticia:
Simple doc that proposes ss13 inspired bitrunners. Alternative to the salvagers and their everlasting out-off-station adventures.
Note that I saw #587 later than this doc was written but its anyway doesn't really touches salvage, so I'm kinda clueless what to do.


kosticia:
I read a couple of comments from roomba in #587 and I can say that this probably against current vision of maints, oops

PR #594, opened 2026-01-30
Add Salvage Post-Mortem to the docs repo.

Princess-Cheeseballs:
## Why?

No reason for it not to be public. I also updated the "current" salvage design doc to say it's been officially retracted per request of mirrorcult. 

I'm going to be using my maints perms to merge this directly into the repo. The PR is here to ensure that I get all the files pushed at the same time. 

A.

### DIFF
--- a/src/SUMMARY.md
+++ b/src/SUMMARY.md
@@ -228,8 +228,9 @@ Space Station 14
 		- [PR Guidelines]()
 
 		- [Proposals]()
-			- [Salvage Proposal](en/space-station-14/departments/cargo/proposals/salvage-proposal.md)
 			- [Mail Deliveries](en/space-station-14/departments/cargo/proposals/mail-delivery.md)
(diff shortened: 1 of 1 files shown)


CollectionOfBones128:
I think this does nail some good aspects!

Giving science a crate full of goodies is an unparalleled feeling, exploring space is amazing, and the VGRoid has unmatched vibes. Plus, the first few times fighting Goliaths was a lot of fun, if the PvE enemies were designed better it'd be even greater!

And yeah, the job board goes against Salvage's set goal (get materials), forcing you to either abandon the task or double up your duties. I wish scrapping and mining instead just gave you progression points you could use to guarantee progression loot if your RNG was bad or something.
And wrecks kneecap the magnet too. Locks it down to a thing that doesn't help anyone.

I hope the amount of knowledge accrued over the last few months can see the rebirth of salvage, there are so many good or at least interesting ideas around.

PR #598, opened 2026-02-08
Rewrite the design document guidelines

ArtisticRoomba:
The design document guidelines have been completely rewritten.

We've been getting a lot of questions as to what a design document *is* and what it's supposed to accomplish. From reviewing old design documents, most are feature lists, which aren't horrible by any means, but they often don't justify why those features are needed and how they contribute to a wider thematic design of whatever someone is adding.

We'd like to push design documents into a tiered structure where previous design documents provide foundation as to why things need to be done, with further design documents establishing how things are done and why it establishes the goals set out by the previous documents. In general, less repetition and derivation of things we already know, and more "how does whatever you want to implement actually align with desired gameplay".

Decorum page was removed because it's unnecessary wording that could be literally anywhere else and it was never followed ever. Game area design document was removed because it's just not where we're pushing docs anymore.

### DIFF
--- a/src/en/general-development/codebase-info/pull-request-guidelines.md
+++ b/src/en/general-development/codebase-info/pull-request-guidelines.md
@@ -15,7 +15,7 @@ Your pull request may be reverted or closed for any reason.
 
 - Please have some familiarity with [C# conventions](https://docs.microsoft.com/en-us/dotnet/csharp/) (if working with C#) and [our own conventions](./conventions.md). Try to read how other parts of the codebase are formatted for a general idea.
 
-- Large new features and comprehensive reworks to existing large features (ie antags or anything that could be considered a subdepartment unto itself), should first be [proposed and accepted in abstract](../feature-proposals.md) before you start working on actually implementing it.
+- Large new features and comprehensive reworks to existing large features (ie antags or anything that could be considered a subdepartment unto itself), should first be [proposed and accepted in abstract](../design-documents.md) before you start working on actually implementing it.
 
 - Read the [Freezes and Restrictions issue](https://github.com/space-wizards/space-station-14/issues/8524) and make sure your PR does not interfere with anything or requires special requirements.
   
--- a/src/en/wizden-staff/maintainer/maintainer-workgroup-policy.md
+++ b/src/en/wizden-staff/maintainer/maintainer-workgroup-policy.md
@@ -106,7 +106,6 @@ Conflicts on pull requests that are internal to the workgroup are subjected to t
 
 ### Design Documents
 Workgroups are encouraged to make design documents for their game area.
-View the [Game Area Design Document Template](../../general-development/game-area-design-doc.md) and the Department Design Document Template on each drafted department for a good document basis.
 Relevant Game Area / Department design documents (which will describe pillars of design for the workgroup's dedicated area) should be merged before a workgroup will be considered fully-formed and functional. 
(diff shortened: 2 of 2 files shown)


src/en/general-development/design-documents.md
```diff
@@ -0,0 +1,126 @@
+# Design Documents and Proposals
+Design documents outline or guide a game area, department, mechanic, or system design.
+They are extremely important for organizing SS14's development.
+
+Overall, a design document seeks to achieve the following goals:
+1. **Design documents are important for establishing a general direction and working towards a vision.**
+   1. It is useful for contributors and maintainers to know whatever you're thinking, and how your feature will fit into the game.
```
RemFexxel:
```suggestion
   1. It is useful for contributors and maintainers to know what you're thinking and how your feature will fit into the game.
```

src/en/general-development/design-documents.md
```diff
@@ -0,0 +1,126 @@
...
+   2. Sometimes things are simply incompatible with a previously existing design or goal, and we don't want you to waste your time making something that has to be rejected at the end of the day. Similarly, you shouldn't write a feature to completion and then attempt to write a design document that justifies your added feature. This usually results in a lot more wasted time.
+   3. It is essential that ideas, problems, solutions, etc. do not get lost to time. SS14 has been burned in the rear by people forgetting *why* we made a decision, going down a path we already found out the hard way as to why it doesn't work. As such, design documents serve as a way to outline your reasoning and preserve your thought process as to why something exists.
```
RemFexxel:
```suggestion
   3. It is essential that ideas, problems, solutions, etc., do not get lost to time. SS14 has been burned in the rear by people forgetting *why* we made a decision, going down a path we already found out the hard way as to why it doesn't work. As such, design documents serve as a way to outline your reasoning and preserve your thought process as to why something exists.
```

src/en/general-development/design-documents.md
```diff
@@ -0,0 +1,126 @@
...
+In this, we've already outlined a few goals:
+- The antagonist adds a spike of chaos while not being extremely problematic to the point where it causes the station a lot of trouble.
+- The pursuer has a single target and must kill it or else it runs out of time. We expect the crew to either intervene or the player to hold a PhD in running away from things.
+- The theming of the antagonist is established - it's a terrifying hunk of metal, and it wants to kill you.
+
+The document then describes a list of features, but makes sure to say why those features are necessary and how they work towards the established design goal.
+
+### Design documents should not delve into technical implementation details
+Design documents seek to establish what your design is, why you're doing what you're doing, and how it should be produced.
+As such, design documents should try to stray away from how a feature will be implemented technically.
```
slarticodefast: The current template has a small section that requests giving a summary of the technical implemanation that will be needed for the feature, which is conflicting with this statement.
We introduced this because we had a bunch of completely unfeasible doc propoposals from authors with no coding experience. I think it makes sense to keep this requirement as part of the doc, because it
- shows maintainers that someone is actually interesting in implementing the proposed feature.
- ensures that the implementation is actually technically feasible rather than wishful thinking and won't cause massive conflicts with existing systems.
- For things like UI and UX suggestions it helps the reader to better conceptualize the new feature as it gives an example of how players will actually interact with the feature, which can vastly differ depending on the way it is finally implemented. This of course does not need to be final or overly detailed, but things like an UI mockup can help a lot.

PR #599, opened 2026-02-10
Plant potency, nutrients, and environmental tolerances rework

Wolfkey-SomeoneElseTookMyUsername:
Copied from the overview:

A rework of the potency, nutrient, water, and environmental tolerance mechanics of botany. Instead of having a potency value that decides the amount of reagents a plant generates, it will instead be decided by the amount of, and type of nutrients in the plant. More powerful reagents require nutrients that are harder to obtain, or have other downsides to their use. Some reagents will also change the environmental requirements of the plant that contains them, such as a cryogenics related reagent causing a plant to require colder temperatures to survive.


Wolfkey-SomeoneElseTookMyUsername:
Should i add a note saying hydroponics trays will automatically give the "water" nutrient to plants they contain? It seems like a good idea, as it reduces the amount of micromanagement, and encourages collaboration with the science department to make new trays instead of just using fungal soil.

Wolfkey-SomeoneElseTookMyUsername:
I have started to implement a version of this on a server downstream. https://github.com/michaelchessall/SS14-Persistence/pull/423

PR #600, opened 2026-02-12
Shipyard updates

Arcticular1:
As the title suggests, this changes the shipyard page - specifically the roster and guidelines


Arcticular1:
OOOPS

PR #601, opened 2026-02-12
Twerg Species Proposal

Lev-tobecontinued:
This proposal is my concept of how to potentially address current issues with base SS14 Dwarves, Please let me know if you have any ideas or suggestions.

image2
image


Djungelskog2:
I really like the sprite work, haven't read the actual doc yet

Djungelskog2:
Just read through, pretty cool and inoffensive but I feel like it doesn't do much unique with them being isopods or being super small. Even if they don't have unique mechanics like leaping or taildrag or rolling up like a pillbug, I think some radicaly different resistances would be a neat change. Did a quick google search and found they are particularly heat resistant but I think the most interesting thing is them essentially having inbuilt armour.

I was thinking they would be especially ballistics and heat resilient because of their adaption to urban and humid 
environments protecting again light heat but particularly their plated backs, providing a massive ballistic and piercing resistance at the cost of them, in regards to being little bugs, extremely vulnerable to any kind of pressure, cold and blunt damage.

bladbochar:
we need this!!

ThanosDeGraf (commented):
Honestly, I like the idea of giving positive effects from metabolizing ethanol to a bug-like race of detrivores, as stuff like rotting fruit famously do contain alcohol, and I understand your reasoning for how this might be a better replacement for dwarves.

One or two mentions of dwarves in your overwiev about the intent is fine, but I'm stumbling over it every paragraph. Consider removing most mentions of it. Besides that I don't see much issue with this doc, though I would like someone else to voice their opinion on it as well.

sowelipililimute (changes requested):
One thing I want to see is leaning more into the shell part of isopods—i.e. there should be a mechanic playing into that. (A unique mechanic, think outside the box of a simple damage modifier.)

Lev-tobecontinued:
okay I am alive, I've updated the document, removing the heat and cold damage bits, and added in a shell based mechanic I think should be fun and also relatively achievable for me to implement given I will have to learn C# for this project

It will take me a while because I'm currently working on Trieste Port related PRs at the moment, but I figured I should at least get the design document squared out.

Dogbone10:
I really like this, like, a LOT.
Delta-V had chitinids which are nice, but these are WEIRD and that's more important.
I'm under the belief that the races in this game should be truly alien in their biology, since if they weren't we'd just be stuck with Florans from Starbound, borgs that are just metal humans, vox being sexy bird furries instead of decrepit space-chickens, etc etc yada yada- and that's not interesting.

BONUS POINTS; These can share sprites with dwarves! It sets a good precedent that the game should be as deep as it is wide mechanically. (Even if it is jut sprites.)

CrackedEmerald:
this is a very fun proposal. I like how they are not only physically different, but physiologically and mechanically as well. 
 Their crab inspiration would give a whole other breadth of different looks for a character, and theyre very different from a distance and especially up close.
 The reagent resistance is very interesting as well, and might give med a preference for cryochems, or even surgery in the future.
Turning into a ball and rolling around sounds really fun, and would be nice to see on station.

Lev-tobecontinued:
Oh by the way, I assume this is fine, I've already gotten some feedback from people who know Nordic culture better than I, specifically for the naming conventions, and specifically how non-binary names are done 

But I do want to make sure I'm using bits of the culture in a respectful way, and if there's anything anyone wants me to lean into, or lean out of. I know we have a lot of players from places like Sweden, Denmark, Norway, and various other countries, so if you are from those cultures, or know much about them, do let me know if there's anything I can do to make this species even better. Most of my understandings are potentially slightly surface level.

PR #603, opened 2026-02-12
Spy Antagonist Design Doc

TheSecondLord:
Started working on this, and figured I should follow guidelines and write a doc before anything is set in stone. Meet the spy.

Never wrote a design doc before, hopefully this is acceptable. Please let me know if any parts need to be rewritten for conciseness because I feel like it gets a little wordy at parts. If you want to discuss on discord, ping me in a development channel @ livelaughlord, or just tell me why it sucks in github comments.

Initial implementation of the bounty board UI, just for examples sake. Subject to change of course.

image


qrwas:
The concept sounds interesting, but there's one thing I don't like: using a single code for every PDA. It sounds... kind of weird?
Maybe it's worth adding some extra interaction that would be accessible to everyone, but at the same time, make the process more engaging.
For example: unscrew the PDA > toggle certain 'contacts' (similar to the access interface that opens when you use a multitool on a door circuit) > and only then enter the code. A sort of 'jailbreak' that downloads malware from the 'bluespace internet.' You could even add an 'installation' delay.
This would slightly slow down the initial setup or recovery, but without knowing which components to interact with, it would prevent accidental discovery—or at least make it highly unlikely. This way, you need to know two things: the current 'jailbreak' sequence and the access code itself.
🤔

TheSecondLord:
> For example: unscrew the PDA > toggle certain 'contacts' (similar to the access interface that opens when you use a multitool on a door circuit) > and only then enter the code.
Some thoughts on this:
- Thematically, I definitely like the idea of having to actually tamper with the PDA, and is something I tried to work into the design but ultimately scrapped.
- Is the idea for the secret code to be inputted through wire pulses (e.g. red - yellow - green - red unlocks the spylink) or is this tampering just a prerequisite before you're able to enter a code? If the former, how does this interact with locking it afterwards? Would cutting these wires have any effect on this? Should these wires be linked to actual functions on the PDA, like disabling it's ability to light up? How would this handle weird interaction priority jank when it comes to chameleon PDAs, which iirc can *also* be locked via screwing?

Djungelskog2:
Thief but good

SolventMercury:
> The idea's good, but if both a wire sequence and a code is required then what's the gameplay reason for needing both? I think it'd be better off with either-or, instead of having 2 codes but each is useless without the other. Also keep in mind that the wires adds an arbitrary requirement for the spy to use their main tool, meaning if you aren't in a role with access to a lathe, you'll have to run to cargo or get lucky in the tools room to get a multitool every time.

Code with no wires: Someone could say the code over chat (especially since code can be learned just from someone else looking at the PDA), now everyone has a spy PDA

Wires with no code: Spies have no way to lock the PDA without also destroying it. Could be desirable for a "This message will explode in 5 seconds" texture to the RP, but also inconvenient.

Code and wires: Spy PDAs can be captured in functional state, but only if a spy is caught off guard, and capturing them only results in at most one PDA in the hands of the spy's enemies.

Effect on TOTs - likely negligible, as the limiting factor to PDA uplink use is access to telecrystals, not just access to uplinks. I forget if tots have personalized codes, but if not, you could theoretically have sec opening uplinks using a code they busted off a different TOT? But I've never seen that happen, so I assume it's either not possible, or so powergamey that even the shittest of secs haven't dared try it.

ScholarNZL:
I had some thoughts on how the unlocking/locking could work:

PDAs now have easily accessible wire terminals, which can be accessed with a barehanded, somewhat lengthy doafter.

Wires are then revealed. PDAs are low voltage, and do not require insuls to tamper with. Certain wires interacted with in a particular order make contact with your spy network / black market trader for a limited time.

Spies get visual cues on this workflow -- non-spies do not.
Each successful tamper will result in a +X% additional chance to break the PDA, which would make the spy need to make wise decisions about when to unlock.

Incorrect tampering by anyone, spy or not, would result in a broken PDA, the effect could be that the pen auto-ejects, or that the screen is cracked, whatever you like, but the intent would be that a tamper would be impossible, and that PDA would need to be replaced.

TechnicViriato:
how the fuck is this still untouched bro

ArtisticRoomba:
breaking news: understaffed maintenance team cannot attend to everything in a semi-timely manner, especially considering recent events.

TheSecondLord:
Made some changes if anyone gaf:
-Removed the stuff about universal PDA access to the bounty board since that's a base feature of uplinks now
-Removed the shared spycode stuff because I thought about it a bit and decided it doesn't really add anything of value.
-Tweaked wording

Some implementation notes:
-Wire pulsing to unlock the board is cool and I'd like it but I'm not including that in base implementation or the doc for now. Copying uplink unlock system wholesale is both much easier and makes barely any difference in terms of gameplay
-The freebies tab would probably be left out of an initial MVP and implemented later
-The only technical parts of this proposal are the stealing system, the black market and the bounty board functionality. Besides that, I'd *very* tentatively say that this would be a comparatively quick and easy antag to implement.

PR #607, opened 2026-02-25
Create Accent Guidelines

SlamBamActionman:
Guidelines for how accents in the game should be. Short and simple.


ThanosDeGraf:
> https://forum.spacestation14.com/t/the-future-of-accents/26642
> 
> We are doing things in reverse order but it _was_ an odd situation and we voted, so. Fine I guess.

image

PR #617, opened 2026-03-16
Rewrote Security's Department Design Document

SolventMercury:
This doesn't precisely adhere to the design doc outline, but I was told the outline wasn't that good anyways. I tried to put more work in on reasoning out the design principles for security. There was a lot of material to cover, so be sure to let me know what I missed.

I know everyone wants docs about stuff like security's arsenal, but I figured I'd start here first.


keronshb (changes requested):

SlamBamActionman:
I have to start out this review with the caveat that being the writer of the existing Sec doc makes me unavoidably biased. 

As-is, I see there being both structural and content issues with this rewrite.

1. There is a lot of casual language used in this document. Admittedly the existing doc's language can be too dry at times (looking back it is concise to the point of reductiveness), but this rewrites features redundant sentences and descriptions. As an example, `The first of security's jobs is unglamorous and rarely sees active discussion, but it is important nonetheless.` serves little to no purpose in a design document and is more suitable for an essay. You can find a good resource here that describes how to format sentences/paragraphs for documentation: https://developers.google.com/tech-writing/one

2. The document puts a lot of describing what Security is, but lacks emphasis on the "why". What is the purpose of Space Law in the game? Why should Security not use lethals all the time? It is critical that a design document can answer these whys because that informs future additions and changes to the department. While reading, there were many moments I felt I could ask "But why?" and it wasn't being answered.

3. Sort of related, I find the document puts emphasis on non-essential parts of Security. Don't get me wrong, contraband handling is part of Security, but nowhere near important enough to be the first meaty paragraph. What Security adds to the game and the round, and the appeal its gameplay should have to players (both Sec and non-Sec) are to me far more important as a design doc.

4. There are contradictory parts in the document, and parts that I think are outright wrong. Security can very well remove contraband from the game - selling contraband is an explicit mechanic, and it is not unheard of to throw bombs etc into space. Security should make the station safe above all else - but also not dish out harsh punishments or mass arrest, which would make it more safe overall. I feel a lot of these could be ironed out by interrogating their purpose in the game.

It may very well be the bias talking but, outside of the existing document being a bit too curt, I don't see this new version improving much in terms of giving guidance when designing new things for the Security department.

PR #629, opened 2026-03-23
test & linter debug guide

### LINKED ISSUE #231: There should be documentation on how to run tests
It's not obvious to most new devs that you have to run both `Content.Tests`, `Content.IntegrationTests` and `Content.YAMLLinter` to properly test your changes. 

There should be a section under "General Development" explaining how to run Tests/IntegrationTests/Linter, both for content and engine. It could also explain how to only run a specific test, using the `--filter` option.

mqole:
resolves #231


src/en/general-development/tips/debugging-tools.md
```diff
@@ -30,11 +30,27 @@ Parameters the `vv` commands take:
...
+
+To run YAML linter, you can use the shell command `dotnet run --project Content.YAMLLinter` in your project folder. If you're using an IDE such as Rider or VSCode, you can also use the YAML linter preset to run your project.
+
+## Running Integration Tests
+
+Integration tests are extremely minimal simulations of what you need to run SS14 - just a server-client pair, no maps or lobbies or anything - and you can run game system code on these simulations in order to locate issues that can only be found through live game simulation. You can use this to make sure a bit of code works as intended by simulating its effects and comparing them against expected values, for example making sure that when you buckle someone to a chair, then that person can't move.
+
+To run integration tests, use the shell command `dotnet test` in your project folder. You can also use `dotnet test --filter` followed by the name of a specific test if you don't want to run every test at once. This is recommended as some tests can take a long time to run! IDEs such as Rider or VSCode may also have built-in ways to run tests without using the terminal.
```
beck-thompson: Could you add a picture like this?

Image

src/en/general-development/tips/debugging-tools.md
```diff
@@ -30,11 +30,27 @@ Parameters the `vv` commands take:
...
+
+When you open a PR to the Wizard's Den repository on GitHub, integration tests will run on your PR automatically. If tests fail, you'll need to update your PR before it can get properly reviewed.
```
beck-thompson: Could you add some info / images on how to see the output of the tests? A lot of new people struggle with this!

beck-thompson (changes requested):
This is amazing! Really necessary part that was missing 🤩

src/en/general-development/tips/debugging-tools.md
```diff
@@ -30,11 +30,27 @@ Parameters the `vv` commands take:
...
+
+```admonish note
+`Content.IntegrationTests` is the directory containing integration tests. It is not the same thing as the similarly-named `Content.Tests` folder. You should run the tests in both of these folders to debug your changes.
```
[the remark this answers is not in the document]
Tayrtahn: The short version is that unit tests test a single "unit" (typically a class) while integration tests test that multiple units work together correctly. 

For example, a there's a unit test that uses spoofed data to make sure that the methods of `LocalizedDataset` work as expected - correctly enumerating its entries, etc. This test doesn't need to bring in any other systems in order to run.

By contrast, there's also an integration test that loads up all the localized dataset instances in the game (going through resource loading, yaml parsing, prototype management, etc) and makes sure that they're set up correctly. 

Unit/integration tests are used widely in software development outside of SS14, so Google can provide you with more info about the concepts and their differences.

Edit: I lied, the "unit" test example actually does use some dependencies, so it's not a "pure" unit test. The hypothetical test I described is still a good example though!

PR #630, opened 2026-03-23
Shipbreaker Salvage - Salvage Rework Proposal (Redux)

SlamBamActionman:
This is a resubmission of a Salvage rework proposal of "Shipbreaker Salvage", colloquially known as BreakSalv. 

It is an overhaul of the original submission https://github.com/space-wizards/docs/pull/575, which was prototyped and playtested internally by staff to determine if it could be a viable redesign for Salvage. The playtests were overall received positively, so this design doc was written to tackle issues raised with the original submission and playtests, promoting further on-station gameplay while still keeping the unique EVA work that Salvage offers. 

Enjoy!


src/en/space-station-14/departments/cargo/proposals/shipbreak-salvage.md
```diff
@@ -0,0 +1,340 @@
...
+
+There are still some cons to working outside the station. It limits crew interactions, as there is no Salvage "front desk" and the Salvage Bays would not be by any high-traffic area (unless a station specifically designs its layout around it). On the other hand, tying Salvage's work to the Salvage Bay will make it massively more accessible compared to previous iterations. Paramedics will have an easier time accessing Salvagers in case of lethal explosions, Cargo & QM will generally know where Salvagers are while working, Engineers can pass by Salvagers when traversing the outside of the station, and there is a much lower barrier of entry for Passengers to help out by virtue of the Salvage Bay being more open.
+
+The Salvage Bay also offers a good spot for holopads, and since the Wreck Timer is purely cosmetic, there is no penalty for Salvagers to pause their work to engage in conversation or complete station tasks.
```
Tayrtahn: What's the Wreck Timer? It's not mentioned in this document. If it refers to the old despawn timer from current salvage gameplay, I would think it should just be removed.

Tayrtahn: Gotcha. Sounds good!

EmoGarbage404 (commented):
I left a comment on the first breaksalv doc so I figured i'd give this one a read as well. 

Generally, the largest sin about this doc is that it's completely unwilling to commit to it's gameplay decisions, leaving huge elements of the proposal as gaping question marks.

The big obvious question that this doc seems unwilling to answer is the question of whether or not wrecks will be mapped, randomly generated, or fuzzed with some combination of the two.

I know the doc says this but like this is not really a statement:

> This document does not make a definitive statement on how the wrecks and wreck types should be implemented technically. There are many options here (some better than others), from handcrafted prefabs to fully randomly generated

For context, issues of premapped wrecks vs. procgen maps was probably one of the single biggest factors in the switch from 2021-era magnet salvage to expeditions as a mechanic. It's been a pretty constant source of maintenance concern and is something which could justify an entire document of its own. It's pretty much indefensible to not actually specify what the plan is for it. FTR i think you could put a good argument for combo mapped + variation passes with spawners but as it stands that's not being made.

The other obvious element which has kinda been left unexplored is this whole notion of "quests" and "relics." I was initially curious when i saw the terms because it seemed like an interesting idea, but looking into the doc, an unbelievable level of detail has not been specified. I'd argue that the doc is borderline unfinished because this extremely important portion of it is just missing a ton of detail.

The doc sorta just poses a bunch of mechanics and outcomes but doesn't really outline how these things would manifest as actual game elements. SS14 is generally a highly immersive and simulated experience, so having something as blatantly "gamey" as an automatic quest tracker just seems like it would be unfitting.

I think there's also a question of maintenance burden when it comes to these quests. Out of the examples given, it appears that every individual quest would need to be individually designed and added to interface with each different job or department in the game. I tried to see how often relics are supposed to appear (every section refers to relics being expounded upon but none of them really do) but at least at a glance it seems like it'd be an obvious weak point where a lack of variation would cause repetitive gameplay.

In general, both the mapping stuff and relic design seems to presuppose that you can just keep making Stuff and gleaning variation and value from it, but from experience with previous salvage systems which had _this same exact problem_, they tend to be pretty much a spiraling cost that doesn't utilize their potential due to not interacting with each other. 

You don't get any combinatorial explosion out of the various salvage elements because all of them are either discrete parts of a gameloop which is itself a fraction of a larger gameloop or are alternate modifiers.

## Misc criticism

### 1.

The whole concept of towing seems really poorly thought out because the doc simultaneously specifies that the grids spawn even closer to the station than the current iteration of the magnet. I don't get why you would need to tow the thing to the station at all.

> Regardless, Salvage will want to move the wreck closer to the Salvage Bay, as it will make it easier to extract materials and is necessary to complete the later steps

It says that this will make it easier, but there isn't anything specified that actually clarifies what these processes are which are made easier by being in the station.

> #### Towing
>
> pulling grids is fun idk what else to say yippee pew pew i'm so strong pulling shottle weee

this is only funny if you've actually presented a coherent argument for the actual thing at hand, fwiw.

### 2.

ok genuinely this is just the 2021 salvage doc but with extra stuff. there were a lot of things about that which didn't work (citation: the 3-5 salvage reworks since then) so i would hope that if this is basically the same thing as that but with additional bells and whistles, there would be some effort put into explaining how those issues were addressed.

sowelipililimute (changes requested):
This document is effectively incomplete in several key areas, while being long-winded and poorly organized.

We do not need an explanation of how an access-locked job area works, lists of modifiers for magnet pulls, etc.

Guidelines are introduced mid-sentence in sections meant to explain the rationale behind the document, and the core of new fundamental mechanics such as relics and questing are obfuscated in lists of what content they should have.

Do/don'ts are shown next to poorly explicated guidelines—rather than being examples that summarize long guidelines or clarify edge cases/design traps, they're often more verbose than the guidelines themselves and leave it to the reader to infer the rest of the guideline picture.

Above all else, I cannot critique a design from a poorly written document, as I don't see the design you have, I see the results and am left to infer the design from those.

Google has good resources on technical writing, which I highly recommend taking the time to study for the next revision of the document: https://developers.google.com/tech-writing

YoungThugSS14:
Hi! Good written document (atleast to me lol) but I was just wondering if debris and asteroids would still spawn in space far out? Would be interesting if a tider could ask HoP to become an explorer so they could loot those

SlamBamActionman:
After playtesting and further design, the _relics_ concept have been replaced with _safes_. They largely follow the same purpose and design, but it constraints the "relic type" down to a single object type (i.e. safes) and also the "relic effect" (the effect is now receiving appropriate "safe loot"), greatly reducing design complexity. The main consideration for their design then becomes the unlock conditions.

This also allows safes to more nicely tie in with Salvage's progression through *magnet chips*, used to provide the salvage magnet with license upgrades. The chips encourage Salvagers to see through the delivery rather than pawning it off to a Cargo Tech. 

More details in the document!

SlamBamActionman:
The document has passed an [approval vote](https://forum.spacestation14.com/t/shipbreaker-salvage-proposal-redux/26968) and as per the [Docs review procedure](https://docs.spacestation14.com/en/wizden-staff/maintainer/doc-review-procedure.html), I'll be marking this PR as approved. This means work can officially begin on implementing this content! 🎉 

The immediate next steps will be to form a workgroup out of interested maintainers (we already have a lineup) and creating a roadmap issue & project on the main repo to track progress. Thank you all for the interest in this gameplay change, and expect to see this document update as playtesting progresses!

BombasterDS2:
Good document and I like it's idea, even if I want to see more interactions with space in space station game. 

But I see big problem with grid splitting charges. Since we can't connect grids, it'll become big huge problem for station crew. Yeah, antags can degrid parts of station even now with RCDs, but it takes huge amount of time and RCD charges. And it's a lot easier to notice since RCD space area after destroying tile. 

Just think that there should be added a way to connect grids in this proposal.

Murrytmds:
Being able to connect grids would be awesome. Its also a feature that, AFAIK, people have tried to add and failed for years.

sowelipililimute (changes requested):
minor style nit while i re-read the new documents, you should have one sentence per line to make future diffs easier to read.

Bokser815:
Salvage was what originally got me invested in Space Station 14 over other projects. I was inspired by the sense of adventure associated with getting control of a shuttle with some new friends every round to supply the station. I had struggled to form any connection with the game before this point. It showed me the potential of shuttles, grids, and planets that made this project so much more unique. I would say that it is in the best interest of the project to continue to build upon such a strong foundation to create unique and exciting gameplay. Upstream should be willing and confident to pioneer new ways to have fun. It will only enrich what downstream is then able to create. This has an accumulating effect as downstream contributors witness these changes; they may even be willing to help keep moving upstream forward. This brings the community and project together as a whole rather than separate niche branches.

SlamBamActionman:
I've updated the Salvage design doc (read: not the rework proposal) with a version primarily written by @sowelipililimute. It features much of the same concepts regarding Salvage's place in the game and its gameplay but written much more succinctly. Enjoy!

PR #638, opened 2026-04-24
Telescience Design Document / Proposal

Mehnix:
Proposal introducing a Telescience Implementation.

I've already made a functional Teleframe implementation some time ago, and a basic Science-gaining prototype. Figure before going any further should get some opinions on it.

This document covers:
- The Teleframe, a structure designed for long-range transportation, alongside its various peripherals.
- Reality Fractures, an invisible entity that produces strange anomalous effects, and where science is gained
- Anomalous Effects, a catchall for the effects made by the Reality Fracture, as well as Teleframes when teleportation goes wrong.

Constructive Criticism is appreciated.

### DIFF
--- a/src/en/space-station-14/departments/science/proposals/telescience.md
+++ b/src/en/space-station-14/departments/science/proposals/telescience.md
@@ -214,7 +214,7 @@ Anomalous Effects have four categories:
 	- Immediately turn the nearest person into a Wizard
  	- Create a portal that after a minute calls forth a large number of dangerous monsters, with a chance of them being zombified.
   	- Creates a highly radioactive item that, after 3 minutes will spawn a singularity. The item can be destroyed in a microwave, sold for a very high price, or thrown into space.
-  	- Create an item that will repeated create lighning, after 3 minutes it spawns a Tesla ball. The item can be destroyed in a micorwave, sold for a very high price, or thrown into space.
+  	- Create an item that will repeated create lightning, after 3 minutes it spawns a Tesla ball. The item can be destroyed in a microwave, sold for a very high price, or thrown into space.
   	- Spawn several new anomalies and reality fractures, after 5 minutes cause every anomaly and reality fracture on the station to supercrit.
   	- Repeatedly spawn major, moderate, and minor anomalous effects for a minute.
   	- Integer overflow the station's cash reserves for one minute.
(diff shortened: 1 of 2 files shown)


miguellemonstre-creator:
Another banger from Mehnix.

src/en/space-station-14/departments/science/proposals/telescience.md
```diff
@@ -0,0 +1,262 @@
...
+This aspect could be further expanded to gain additional science from cataloguing these effects, however this is outside the scope of Telescience. Anomalous shards could be found by other means, such as through salvaging.
+
+> John Medical Intern is handed an Anomalous Shard discovered by the Science Team alongside a bottle of Artifact Glue and a Screwdriver. Every time he uses the shard, its pulse aids in the healing process of his patients, albiet blinding them temporarily. Whenever he runs out of charges, he just needs a splash of Artifact Glue to fix it up, then a Screwdriver to reactivate it. 
+
+## Design Intentions Summary
+In conclusion, these are the primary Pillars and Intentions by which Telescience is intended to be implmented.
+
+Crew-side Teleframes for use in Telescience should embody the following:
+- Provide a powerful yet limited form of traversal. The Player should not be able to teleport into a restricted area round start unhindered. They should however be able to use their tools and skills to teleport into these places with time spent preparing to do so. (Such as extrapolating coordinates to the captain's room from those nearby).
```
ketufaispikinut, line 236: Extrapolating coordinates won't really be that complicated. Could teleframe coordinates be non-linear, perhaps?

Mehnix: Space Station 13 has something like this, although this has the issue of 1: being pretty unintuitive, and 2: at most being a one-time inconvenience at the start of the shift until you figure out what the modifier is.

It existed in 13 I think because coordinates and orientation were static, so you could put in the same coordinates each shift to go to the same location if it weren't for a randomly assigned non-linear modifier on X and Y positions. As this does vary in 14 the limit is instead you have to physically go to the place you want to steal from, stand around for a while to get the right coordinates, and only then be able to teleport. The teleframe console says over radio exactly where the target of a teleportation is including referencing map beacons so it'll be immediately obvious what you're doing if it says "nearby the captain's room". The effect itself is loud enough anyone nearby will see or hear it.

If it ever became an issue that people were speedrunning yoinking something important it wouldn't be too difficult to add though. Although at most it'd delay by a minute or two.

Also anything anchored can't be teleported, so I feel the only item really at risk is the captain's spare ID or maybe head lockers which are unanchored on most maps. Otherwise you'd have to teleport in to an area to get the item, and then have another way to get out.

ketufaispikinut: By non-linear I mean't more like distorted by a vector field :P.

The reasoning agains't it makes sense though (altough people are already rushing the captain's spare ID in the first instants of the shift, so this should be kept in mind).

PR #640, opened 2026-04-28
Solari Species Doc

spiritaaa:
This document is an introduction to the Solari race (Name pending) as a round-start species. 

I'm extremely open to feedback, I still think a lot of the ideas here are pretty rough. Suggestions are more then welcome.

An example of what they could look like.
SolariSample v3


YoungThugSS14 (commented):
This document should be split into complete paragraphs rather than lines that cut off, makes it easier to review

PR #644, opened 2026-05-04
update bikehorn guide to modern standards

slarticodefast:
A lot of stuff referenced was outdated.

Also added namespace imports since I've seen that confuse a lot of new coders.

### DIFF
--- a/src/en/ss14-by-example/adding-a-simple-bikehorn.md
+++ b/src/en/ss14-by-example/adding-a-simple-bikehorn.md
@@ -328,7 +328,7 @@ Also, `PlayPvs` automaticly manages distance filtering, so you don't have to wor
 With that, this tutorial is finished! If you want to continue experimenting with your newfound clown horn, here are some ideas:
 
 - Try to implement clown horn using existing components. You can refer to skub.yml up this page.
-- Add a cooldown to limit the rate the sound can be used by adding `UseDelayComponent` to your prototype.
+- Add a cooldown to limit the rate the sound can be played at by adding `UseDelayComponent` to your prototype.
 - Adjust the volume/variation of the sound played (see the `PlayPvs()` function's `audioParams` argumernt).
 - Make the sound play when the bike horn is stepped on as well.
     - This one is kind of hard and involves adding a lot of new data! Look at glass shards for an example.
--- a/src/en/ss14-by-example/adding-a-simple-bikehorn.md
+++ b/src/en/ss14-by-example/adding-a-simple-bikehorn.md
@@ -1,18 +1,16 @@
 # Adding a Simple Bike Horn
 
-{{#template ../templates/outdated.md}}
+This tutorial goes over the **Entity Component System** pattern and several other key topics in the SS14 codebase by demonstrating how one would implement a clown horn from scratch. You can try copying the steps yourselves, or you can just read along.
 
-This tutorial goes over the **entity component system** system and several other key topics in the SS14 codebase by demonstrating how one would implement a clown horn from scratch. You can try copying the steps yourselves, or you can just read along.
+## Entities, Components, and Systems
 
-## Entities, components, and systems
-
-While Space Station 14 is written in C#, an object-oriented programming language, it uses a different data model to represent items in game. This data model is called the *entity component system* (ECS). (*Why do we do this? See [ECS](../robust-toolbox/ecs.md)*)
+While Space Station 14 is written in C#, an object-oriented programming language, it uses a different data model to represent items in game. This data model is called the *Entity Component System* (ECS). (*Why do we do this? See [ECS](../robust-toolbox/ecs.md)*)
 
 ### Entities
 
-Each item in game is represented by an *entity*. Players, bananas, stun batons, are all represented by entities. An entity is represented by an integer. No two entities share the same integer representation.
+Each object in-game is represented by an *entity*. Players, bananas, stun batons, audio sources, or even more abstract concepts like game rules or objectives are all represented by entities. An entity is represented by an integer. No two entities share the same integer representation.
 
-By themselves, entities only distinguish one item from another. Without any components, an entity has no behavior.
+By themselves, entities only distinguish one item from another. Without any components an entity has no behavior.
 
 ### Components
 
@@ -20,7 +18,7 @@ By themselves, entities only distinguish one item from another. Without any comp
 
 1. **Label specific entities as having specific behavior.** For example, in one particular game, the entity represented by the integer 37629 contains a `NukeComponent` and a `ActivatableUIComponent`. This means that this entity behaves like a nuke, and also has a user interface that can be raised by activating it.
 
-2. **Store data required to process its behavior.** For example, a `NukeComponent` may have a data field `Timer` that represents how much time is left until the nuke detonates.
+2. **Store data required to process its behavior.** For example, a `NukeComponent` may have a datafield `Timer` that represents how much time is left until the nuke detonates.
 
 Still, components do not contain any logic for processing this behavior. Behaviors are implemented in entity systems.
 
@@ -30,7 +28,7 @@ An *entity system* (often abbreviated to "system") contains logic that implement
 
 Entity systems implement behavior by defining *event handlers* or by implementing a per-tick *update* method.
 
-As an another example, consider the `FoodComponent`. A programmer might make `EatingSystem` to handle eating food. `EatingSystem` listens to the `OnUseInHand` event - whenever `OnUseInHand` is heard/triggered, `EatingSystem` checks if there is a `FoodComponent` in the object that was used. If there is, then it lowers the value of `nutritionLeft` and plays a munching sound.
+As an another example, consider the `EdibleComponent`. A programmer might make `IngestionSystem` to handle eating food. `IngestionSystem` listens to the `UseInHandEvent` - whenever `UseInHandEvent` is heard/triggered, `IngestionSystem` checks if there is a `EdibleComponent` in the object that was used. If there is, then it lowers the value of the nutrition inside the entity and plays a munching sound.
 
 That's the jist of ECS. If you're interested in learning more about it, then check out [Your mind on ECS](../robust-toolbox/ecs.md). The ECS approach really is powerful and allows us to avoid spaghetti code, despite the complexity of SS14.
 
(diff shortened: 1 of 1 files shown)


src/en/ss14-by-example/adding-a-simple-bikehorn.md
```diff
@@ -30,7 +28,7 @@ An *entity system* (often abbreviated to "system") contains logic that implement
 
 Entity systems implement behavior by defining *event handlers* or by implementing a per-tick *update* method.
 
-As an another example, consider the `FoodComponent`. A programmer might make `EatingSystem` to handle eating food. `EatingSystem` listens to the `OnUseInHand` event - whenever `OnUseInHand` is heard/triggered, `EatingSystem` checks if there is a `FoodComponent` in the object that was used. If there is, then it lowers the value of `nutritionLeft` and plays a munching sound.
+As an another example, consider the `EdibleComponent`. A programmer might make `IngestionSystem` to handle eating food. `IngestionSystem` listens to the `UseInHandEvent` - whenever `UseInHandEvent` is heard/triggered, `IngestionSystem` checks if there is a `EdibleComponent` in the object that was used. If there is, then it lowers the value of the nutrition inside the entity and plays a munching sound.
 
 That's the jist of ECS. If you're interested in learning more about it, then check out [Your mind on ECS](../robust-toolbox/ecs.md). The ECS approach really is powerful and allows us to avoid spaghetti code, despite the complexity of SS14.
```
kontakt:
```suggestion
That's the gist of ECS. If you're interested in learning more about it, then check out [Your mind on ECS](../robust-toolbox/ecs.md). The ECS approach really is powerful and allows us to avoid spaghetti code, despite the complexity of SS14.
```

src/en/ss14-by-example/adding-a-simple-bikehorn.md
```diff
@@ -1,26 +1,24 @@
...
+While Space Station 14 is written in C#, an object-oriented programming language, it uses a different data model to represent items in game. This data model is called the *Entity Component System* (ECS). (*Why do we do this? See [ECS](../robust-toolbox/ecs.md)*)
 
 ### Entities
 
-Each item in game is represented by an *entity*. Players, bananas, stun batons, are all represented by entities. An entity is represented by an integer. No two entities share the same integer representation.
+Each object in-game is represented by an *entity*. Players, bananas, stun batons, audio sources, or even more abstract concepts like game rules or objectives are all represented by entities. An entity is represented by an integer. No two entities share the same integer representation.
 
-By themselves, entities only distinguish one item from another. Without any components, an entity has no behavior.
+By themselves, entities only distinguish one item from another. Without any components an entity has no behavior.
```
RemFexxel:
```suggestion
By themselves, entities only distinguish one item from another. Without any components, an entity has no behavior.
```

src/en/ss14-by-example/adding-a-simple-bikehorn.md
```diff
@@ -185,7 +183,7 @@ You must remove the `Component` part of the class suffix when using them in the
 
 Well, this is boring; not only does our component not have any data, but it doesn't do anything either!
 
-Let's add some data to our component. As you may have noticed above, the `Sprite` component on our bike horn has two fields listed: `sprite`, and `state`. Whatever you put in these fields will be passed into the component when it's created, and then our EntitySystem can use that data to do something.
+Let's add some data to our component. As you may have noticed above, the `Sprite` component on our bike horn has two fields listed: `sprite`, and `state`. Whatever you put in these fields will be passed into the component when it's created, and then our `EntitySystem` can use that data to do something.
```
RemFexxel:
```suggestion
Let's add some data to our component. As you may have noticed above, the `Sprite` component on our bike horn has two fields listed: `sprite` and `state`. Whatever you put in these fields will be passed into the component when it's created, and then our `EntitySystem` can use that data to do something.
```

src/en/ss14-by-example/adding-a-simple-bikehorn.md
```diff
@@ -220,31 +218,35 @@ All you need to do to create a field that can be modified in YAML is to add the
 
 Now we're getting somewhere! One thing to note is that the path here is relative to the `Resources` directory (which `SoundSystem` always assumes), and we're also assuming that the `Resources/Audio/Items/bikehorn.ogg` file is real. If you check, it is! But if a sound isn't present that you need, you can always add it yourself somewhere in the `Audio` folder.
 
+Note that the name of the datafield was automatically converted from the PascalCase `Sound` when referenced in C# into the camelCase `sound` when defined in yaml.
```
RemFexxel:
```suggestion
Note that the name of the datafield was automatically converted from the PascalCase `Sound` when referenced in C# into the camelCase `sound` when defined in YAML.
```

RemFexxel (commented):
Nitpicks, but this overall looks fine to me. 👍

src/en/ss14-by-example/adding-a-simple-bikehorn.md
```diff
@@ -1,16 +1,14 @@
...
 
-## Entities, components, and systems
-
-While Space Station 14 is written in C#, an object-oriented programming language, it uses a different data model to represent items in game. This data model is called the *entity component system* (ECS). (*Why do we do this? See [ECS](../robust-toolbox/ecs.md)*)
+While Space Station 14 is written in C#, an object-oriented programming language, it uses a different data model to represent items in game. This data model is called the *Entity Component System* (ECS). (*Why do we do this? See [ECS](../robust-toolbox/ecs.md)*)
 
 ### Entities
 
-Each item in game is represented by an *entity*. Players, bananas, stun batons, are all represented by entities. An entity is represented by an integer. No two entities share the same integer representation.
+Each object in-game is represented by an *entity*. Players, bananas, stun batons, audio sources, or even more abstract concepts like game rules or objectives are all represented by entities. An entity is represented by an integer. No two entities share the same integer representation.
```
Fildrance, line 11: i'd propose to change wording about ids a little bit:

> Each entity upon creation is automatically assigned a unique integer identifier, and no two entities can share same. 

I do understand that talking about NetEntityId and EntityUid here is not good idea, thats for sure. So i guess even mentioning EntityUid, which we use most of the time, would probably be incorrect? Not sure. But if you think that its okay to mention it here - we probably can mention it as something like that (probably wrap in separate panel or expander )

> Each entity upon creation is automatically assigned a unique integer identifier, and no two entities can share same. In code we usually reference game objects by id using EntityUid type, or wrappers like Entity<...> which include such id.

| ℹ️ **Info** |
| :--- |
| With client-server approach comes networking problems, details about that are described on separate page (https://docs.spacestation14.com/en/robust-toolbox/netcode/net-entities.html?) and are more advanced topic, not required for understanding to finish this guide.  |

src/en/ss14-by-example/adding-a-simple-bikehorn.md
```diff
@@ -30,9 +28,9 @@ An *entity system* (often abbreviated to "system") contains logic that implement
 
 Entity systems implement behavior by defining *event handlers* or by implementing a per-tick *update* method.
 
-As an another example, consider the `FoodComponent`. A programmer might make `EatingSystem` to handle eating food. `EatingSystem` listens to the `OnUseInHand` event - whenever `OnUseInHand` is heard/triggered, `EatingSystem` checks if there is a `FoodComponent` in the object that was used. If there is, then it lowers the value of `nutritionLeft` and plays a munching sound.
+As an another example, consider the `EdibleComponent`. A programmer might make `IngestionSystem` to handle eating food. `IngestionSystem` listens to the `UseInHandEvent` - whenever `UseInHandEvent` is heard/triggered, `IngestionSystem` checks if there is a `EdibleComponent` in the object that was used. If there is, then it lowers the value of the nutrition inside the entity and plays a munching sound.
```
Fildrance, line 31: Probably worth rephrasing last part by saying
> then it lowers the value of the nutrition inside the entity (by changing values on EdibleComponent that is attached to it), and plays a munching sound.

I am not 100% sold on this, because clean-li-ness is quite critical for beginner guide, and additional complexity by this phrase in braces could be not justified, but thats my proposal at least.

src/en/ss14-by-example/adding-a-simple-bikehorn.md
Fildrance, line 43: probably worth detalization 

> Entity prototypes define *which components are on the entity upon initialization, and what data those components hold at initialization time*.

Although i am not sure on using word 'initialization' because we have such step in component lifetime and it could be misunderstood that it is only works while component is in initialization stage? But that doesnt sound like a beginner mistake.

src/en/ss14-by-example/adding-a-simple-bikehorn.md
```diff
@@ -202,7 +200,7 @@ public sealed partial class PlaySoundOnUseComponent : Component
 }
 ```
 
-All you need to do to create a field that can be modified in YAML is to add the `[DataField]` attribute, which holds the name of the field, and give it a default value, in this case `string.Empty`. Now, we can add our sound to our bike horn prototype:
+All you need to do to create a field that can be modified in YAML is to add the `[DataField]` attribute, and give it a default value, in this case `string.Empty`. Now, we can add our sound to our bike horn prototype:
```
Fildrance, line 203: I am not sure its valuable to say about string.Empty here - maybe omit it?

src/en/ss14-by-example/adding-a-simple-bikehorn.md
```diff
@@ -220,31 +218,35 @@ All you need to do to create a field that can be modified in YAML is to add the
...
-    
+
 public sealed class PlaySoundOnUseSystem : EntitySystem
 {
 
 }
 ```
 
-You'll notice that here, our system inherits from `EntitySystem`. This automatically registers it as a proper EntitySystem in the game and allows us to use some useful dependencies and override some methods to add behavior.
+You'll notice that here, our system inherits from `EntitySystem`. This automatically registers it as a proper `EntitySystem` in the game and allows us to use some useful dependencies and override some methods to add behavior.
```
Fildrance, line 239: there is a small mess of using 'code' formatting:
in this line it is used now for both existing in code type EntitySystem - which is correct,
and for term EntitySystem - which is a bit odd.

Fildrance (commented):
Few ideas for improvements.
Used guide, it works, new class references are valid.
I propose fixing stuff i found and merging, then asking peeps triagers to try collect feedback on it from newcomers.

src/en/ss14-by-example/adding-a-simple-bikehorn.md
```diff
@@ -267,16 +269,22 @@ You've probably noticed that this code actually gives you an error, *because the
...
+// Content.Server/Sound/PlaySoundOnUseSystem.cs
+
+using Content.Shared.Interaction.Events;
+using Robust.Shared.Audio.Systems;
+
 namespace Content.Server.Sound;
 
 public sealed class PlaySoundOnUseSystem : EntitySystem
 {
     [Dependency] private readonly SharedAudioSystem _audio = default!;
```
slarticodefast, line 286: reminder for myself: As per the newest engine release this should not be readonly and the System needs to be partial
I should also check all other docs for this

src/en/ss14-by-example/adding-a-simple-bikehorn.md
```diff
@@ -267,16 +269,22 @@ You've probably noticed that this code actually gives you an error, *because the
...
-    
+
     public override void Initialize()
     {
         SubscribeLocalEvent<PlaySoundOnUseComponent, UseInHandEvent>(OnUseInHand);
```
Fildrance, line 290: now needs an update to use attribute

PR #646, opened 2026-05-09
Design Doc to Remove Microwave Cooking

Ewanderer:
I had the idea to work on cooking in my head ever since I saw the freeze and open invitation to do stuff. But before I work on such a major overhaul, I figured it would be best to layout my thoughts and design idea for approval both in terms of gameplay but also engine considerations.


Ewanderer:
started prototyping, and i realized i needed a bunch of changes to the originald esign. so no its a draft until i have a prototype i can document.

PR #515, opened 2025-09-12
Cyprin Roundstart Species

TrixxedHeart:
Cyprin are inspired by downstream Thaven, going into a more focused direction of being space carp/dragon people.

They’re a fully aquatic-inspired species that emphasize roleplay through Resonances, a system that guides personality and behavior without mechanical enforcement. Effectively acting as roleplay prompts.
CyprinConcept
Cyprin are humanoid descendants of ancient bio-engineering experiments that merged humanoid, carp, and draconic traits. Their defining feature is Resonance, an internal “frequency” that shapes their beliefs and compulsions, encouraging character-driven conflict and storytelling.

Mechanically, they feature distinct atmospheric interactions and a stun-based punch that sets them apart in gameplay.

Conceptually, Resonances are something I feel will be a major point of discussion. I've been very careful as to try and curate the Resonance list as much as possible and ensure clear messaging for the player that Resonances are something that should be thoughtfully acted upon. I'm happy to adjust Resonances as seen fit but I see this as quite possibly their most important feature, it gives them a unique identity and a fun twist for people who want to roleplay but aren't sure what to do, and for people who want to just have fun playing into different vibes every round.

PR #655, opened 2026-05-28
Cyprin Roundstart Species

chortpentimus:
This is a continuation of https://github.com/space-wizards/docs/pull/515 by @TrixxedHeart

## Original description:
Cyprin are inspired by downstream Thaven, going into a more focused direction of being space carp/dragon people.

They’re a fully aquatic-inspired species that emphasize roleplay through Resonances, a system that guides personality and behavior without mechanical enforcement. Effectively acting as roleplay prompts.
image
Cyprin are humanoid descendants of ancient bio-engineering experiments that merged humanoid, carp, and draconic traits. Their defining feature is Resonance, an internal “frequency” that shapes their beliefs and compulsions, encouraging character-driven conflict and storytelling.

Mechanically, they feature distinct atmospheric interactions and a stun-based punch that sets them apart in gameplay.

Conceptually, Resonances are something I feel will be a major point of discussion. I've been very careful as to try and curate the Resonance list as much as possible and ensure clear messaging for the player that Resonances are something that should be thoughtfully acted upon. I'm happy to adjust Resonances as seen fit but I see this as quite possibly their most important feature, it gives them a unique identity and a fun twist for people who want to roleplay but aren't sure what to do, and for people who want to just have fun playing into different vibes every round.


Hitlinemoss:
Two quick thoughts from reading this:
- Resonances being a pure roleplay mechanic without mechanical enforcement _or_ rules enforcement seems like it'd kind of fall apart in an LRP environment (where people would just ignore them entirely), and even in an MRP environment I think players would prefer to have full control over their character's beliefs rather than having them assigned randomly. While I don't hate the concept from a thematic standpoint, it's important to remember that in a game like SS14 players have their own ideas about their character's beliefs (and their character's species), so forcing randomized RP prompts on them won't go over well, especially when there's no punishment for just outright ignoring them. The whole mechanic also just results in more work for admins in multiple ways, and I also think a lot of players would be frustrated by wanting to play the cool dragon species but not wanting to deal with the forced RP prompts, so I'd prefer to just not have them. (That being said, I'm the only one to respond to this design doc so far so it'd probably be more valuable to have actual maintainers/admins weigh in on this.)
- The association with electricity is a bit odd given that that's not really a power that actual dragons/carps have. That's not to say that I don't like the idea of a playable species with an electrical theme, it just feels a bit out-of-place here?

chortpentimus:
I do agree that the association with electricity is out of place and that the stun punches are a bit redundant due everyone being able to shove. I thought about changing or removing it from the doc, but I figured I'd wait for a second opinion. 

The way that I believe Resonances are supposed to be interpreted is more as an urge or suggestion rather than a belief of the character and that the player is free to resist or embrace them depending on what they want for their roleplay. For me personally, the upside of having an extra gateway into roleplay and encouragement to do things your character wouldn't normally do outweighs the downside of losing a small amount of freedom with how you engage in roleplay for the round. Considering that playing silicons is much more restrictive as following your laws is not optional, I don't see it as a major flaw. 

Maybe, for players that don't wish to interact with resonances at all, it would be beneficial to have a trait that hides them? Thematically, this could be a standard implant made for Cyprin to block resonances. I think this should be satisfactory for people that want to play with or without them, or alternate between both options.

In terms of work for admins, I can't think of anything right now other than making sure there is clear instruction that resonances are not an excuse for breaking rules, writing resonances in a way that avoids encouraging rule-breaking, or (and hopefully this is not necessary) playtime restrictions to prevent new players that may not fully understand the rules from using resonances.

PR #656, opened 2026-05-30
EVA Storage design doc

Hitlinemoss:
This doc details the purpose of the EVA Storage room, how it should generally be designed, and what items it should typically contain.

Consider this a companion to https://github.com/space-wizards/space-station-14/pull/41416. Despite EVA Storage being mapped on _every_ station, it's frustratingly inconsistent across stations. Some stations let anyone with Externals go into it, some are Command-only, some are Command _or_ External, some are _only the Head of Personnel specifically_. Accesses aside, there's also other weird consistencies, like Box having weird departmentally-locked closets containing single hardsuits and Oasis including full jetpacks _and_ grappling hooks. I think if we're going to map this room on every station we should have a consistent guideline for how it should be mapped.

This is my first time making a design doc so feedback is appreciated.

### DIFF
--- a/src/SUMMARY.md
+++ b/src/SUMMARY.md
@@ -168,6 +168,8 @@ Space Station 14
 	- [PR Guidelines](en/space-station-14/mapping/guidelines.md)
 	
 	- [Dungeons](en/space-station-14/mapping/dungeons.md)
+
+	- [EVA Storage](en/space-station-14/mapping/eva-storage.md)
 	
 	- [Guides]()
 		- [General Guide](en/space-station-14/mapping/guides/general-guide.md)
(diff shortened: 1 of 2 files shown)


K-Dynamic:
Centralized EVA storage is a half-baked concept given:
1. It's almost never located anywhere near an external airlock
2. It can't be spread across multiple external airlocks without decentralization
3. It competes with emergency EVA and department hardsuits

Hitlinemoss:
> 1. It's almost never located anywhere near an external airlock 
> 2. It can't be spread across multiple external airlocks without decentralization

I don't think "EVA Storage isn't near exterior airlocks" is necessarily a problem? Though I could see the convenience of stations instead having their exterior airlocks incorporate a "mini" EVA Storage (i.e 1-2 suit storages, oxygen/nitrogen cans, maybe magboots/mini-jetpacks). That'd require a lot of mapping changes to implement, though; adjusting the contents of the existing EVA Storage rooms would be much easier to implement. I think the centralized system also has the advantage of making it easier to tell when a theft happens; if "good" EVAs (rather than emergency EVAs) were something you could grab from the random airlock buried in the depths of maints, there'd be no risk to stealing them.

> 3. It competes with emergency EVA and department hardsuits

This is mostly a problem of emergency EVAs being absolutely everywhere on the station + being basically as good for regular crew (who don't mind the speed penalty as much as, like, security would).

Emisse:
looks good man

Hitlinemoss:
> looks good man

@Emisse slight clarification I'd like from you specifically: this design doc says that EVA Storage should be Command/External access, but you've said regarding [#45219](https://github.com/space-wizards/space-station-14/pull/45219) that you'd like EVA Storage to be HoP access. Which is preferred?

I'll note that my preference would be for Command/External for the reasons mentioned in the current version of this design doc, plus most maps already use one or the other anyway so it's less disruptive to existing maps to change it to use either.

PR #657, opened 2026-06-06
Centcomm intern proposal

n1-nbt:
### Central Command Intern
Ever played as HoP and thought, "WoW I love this paperwork! I wish I wasn't on this chaotic station so I could focus on my paperwork more!"
If you have, Centcomm is looking for ~~drones~~ \*eager workers\* just like you! You can just sit in Centcomm all day and do nothing but paperwork!

Anyways, this is a [genuine proposal](https://github.com/n1-nbt/ss14-docs/blob/cc-intern-proposal/src/en/space-station-14/round-flow/proposals/centcomm-intern.md) for a role I think has good potential if done right, and I would love some feedback and discussion.

### DIFF
--- a/src/SUMMARY.md
+++ b/src/SUMMARY.md
@@ -213,6 +213,7 @@ Space Station 14
 		- [Revolutionaries Rework](en/space-station-14/round-flow/proposals/revolutionaries-codeword-rework.md)
 		- [Tourists](en/space-station-14/round-flow/proposals/tourists.md)
 		- [Station Ecology](en/space-station-14/round-flow/proposals/station-ecosystem.md)
+		- [Central Command Intern](en/space-station-14/round-flow/proposals/centcomm-intern.md)
 		
 - [User Interface](en/space-station-14/user-interface.md)
 	- [PR Guidelines]()
--- a/src/en/space-station-14/round-flow/proposals/centcomm-intern.md
+++ b/src/en/space-station-14/round-flow/proposals/centcomm-intern.md
@@ -1,0 +1,132 @@
+# Central Command Intern
+
+| Designers | Coders | Implemented | GitHub Links |
+|------------|------------|-------------|--------------|
+| n1-nbt | n1-nbt? | :x: No | TBD |
+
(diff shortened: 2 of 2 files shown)


Hitlinemoss:
As someone who likes roleplaying boring bureaucratic jobs, this interests me. Thoughts from reading the design doc:
- One concern I have with giving the CC intern a camera feed of the station (even just via computer) is that it'd kind of end up playing like a "discount AI", in the sense that it'd spend a lot of downtime remotely observing the station. Even something like a station map would allow them to spot players breaking down the walls to heads' offices and whatnot. Unlike AI proper, it'd also not be bound by silicon laws, so the intern has no reason to not just immediately report any crimes it witnesses to the station via fax or announcement console or something. This isn't necessarily a game-breaking balance issue, but it'd be annoying for antags to deal with (because now there's effectively two AIs that could be watching the station any given time) and also just doesn't fit CC's theme (CC generally seems to be pretty hands-off with the station's day-to-day). That being said, not having a camera feed also makes the intern's awareness of the station entirely based on messages manually sent from the station, which could leave the role with not much to do (especially in an LRP environment where Command might not care about the bureaucracy).
- Giving the CC Intern the ability to send ERT/CBURN without admin intervention would be something we'd need to make balance adjustments for. This isn't necessarily a "problem" (it's more interesting for the nukies to need to fight off the ERT once the majority of the station is dead rather than just waiting five minutes to win), but it does add another obstacle that these antagonists would come across reasonably often. At the same time, we'd also need to make sure rounds remain balanced when there _isn't_ an intern and these extra ghostroles never get called; this still isn't necessarily a problem (the Captain or AI being absent also have significant impact on rounds, for instance) but it's still something to keep in mind, especially when (unlike station roles) there's no way to "promote" an on-station player to CC Intern mid-round.
- Minor tidbit, but command gear probably shouldn't be purchasable with the CC budget either, since if those items do get stole ideally Security should work towards finding and recovering them instead of just going "well whatever we can just buy a new digiboard".

PR #669, opened 2026-08-17
The chosen

Velken:
EDIT: moved to outline, check there fore most recent up to date version: https://outline.spacestation14.com/s/21d0b38b-4ac1-4ed8-abfa-5a180b1b194c

Mechanically fixing the Throngler to reduce admin workload.
TL;DR:
Using Throngler turns you into an antag that is valid for everyone else, adds way for crew to destroy it.

### DIFF
--- a/src/SUMMARY.md
+++ b/src/SUMMARY.md
@@ -211,6 +211,7 @@ Space Station 14
 		- [Changeling](en/space-station-14/round-flow/proposals/changeling.md)
 		- [Paradox Clone](en/space-station-14/round-flow/proposals/paradox-clone.md)
 		- [Revolutionaries Rework](en/space-station-14/round-flow/proposals/revolutionaries-codeword-rework.md)
+		- [The Chosen](en/space-station-14/round-flow/proposals/the-chosen.md)
 		- [Tourists](en/space-station-14/round-flow/proposals/tourists.md)
 		- [Station Ecology](en/space-station-14/round-flow/proposals/station-ecosystem.md)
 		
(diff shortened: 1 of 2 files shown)


Hitlinemoss:
I'm not huge on the Throngler always immediately teleporting to a random place on the station as soon as it's spawned; feels too video-gamey. I think if you open a lottery crate and get the Throngler in it, the Throngler should actually _be in the crate_ instead of just arbitrarily deciding to show up in Medical for some reason the exact moment the crate was opened.

Lottery aside, I also just think that there's more interesting ways to spawn the Throngler than just "pops into existence somewhere random on the station" (which isn't to say that should _never_ happen, but I feel like that's more appropriate as, e.g, the result of an artifact trigger). For instance, maybe the Throngler could slam into the station similarly to a meteor / immovable rod, or maybe Salvage finds it on a wreck, or maybe dragons sometimes spawn with the skeletons of past victims inside their stomachs and one of those skeletons can rarely have the Throngler.

Likewise, teleporting _everyone on the station_ somewhere random feels too disruptive to whatever else was going on that round up to that point.

Velken:
> I'm not huge on the Throngler always immediately teleporting to a random place on the station as soon as it's spawned; feels too video-gamey. I think if you open a lottery crate and get the Throngler in it, the Throngler should actually _be in the crate_ instead of just arbitrarily deciding to show up in Medical for some reason the exact moment the crate was opened.
> 
> Lottery aside, I also just think that there's more interesting ways to spawn the Throngler than just "pops into existence somewhere random on the station" (which isn't to say that should _never_ happen, but I feel like that's more appropriate as, e.g, the result of an artifact trigger). For instance, maybe the Throngler could slam into the station similarly to a meteor / immovable rod, or maybe Salvage finds it on a wreck, or maybe dragons sometimes spawn with the skeletons of past victims inside their stomachs and one of those skeletons can rarely have the Throngler.
> 
> Likewise, teleporting _everyone on the station_ somewhere random feels too disruptive to whatever else was going on that round up to that point.

The mechanism to how the Throngler gets to the station isn't something set on stone. The doc is a general direction compass, not a recipe book.

However, the Throngler from lottery will NEVER remain on ATS on spawn, and this is set in stone. (in sum: should NOT be 'cargo exclusive/priority')

PR #671, opened 2026-08-19
Update conventions for shared change

metalgearsloth:
Docs repo has a lot of fluff so not sure where else is relevant.


Princess-Cheeseballs (approved):

Fildrance:
As stated in maints meeting - we are not chaning all of them in one go (sadly) but we do not merge new stuff that uses old naming scheme. :pepeHands:.

